PR #1, opened 2024-06-04
Create Splash Screen

sdelator:
### **Newly Added:** 
create splash screen; added tasky minmap; added dependencies; UI for login and register screen

### **UI Updates**
image
image
image
image

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -47,7 +47,7 @@ android {
 }
 
 dependencies {
-
+    implementation(libs.androidx.core.splashscreen)
     implementation(libs.androidx.core.ktx)
     implementation(libs.androidx.lifecycle.runtime.ktx)
     implementation(libs.androidx.activity.compose)
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -52,22 +52,22 @@ android {
 
 dependencies {
     implementation(libs.androidx.core.splashscreen)
-    implementation("androidx.core:core-ktx:1.10.1")
-    implementation("androidx.lifecycle:lifecycle-runtime-ktx:2.6.1")
-    implementation("androidx.activity:activity-compose:1.9.0")
-    implementation("androidx.compose.ui:ui:1.4.3")
-    implementation("androidx.compose.ui:ui-tooling-preview:1.4.3")
-    implementation("androidx.compose.material3:material3:1.1.1")
+    implementation(libs.androidx.core.ktx)
+    implementation(libs.androidx.lifecycle.runtime.ktx)
+    implementation(libs.androidx.activity.compose)
+    implementation(libs.ui)
+    implementation(libs.ui.tooling.preview)
+    implementation(libs.material3)
 
     // Hilt dependencies
-    implementation("com.google.dagger:hilt-android:2.44")
-    kapt("com.google.dagger:hilt-compiler:2.44")
+    implementation(libs.hilt.android)
+    kapt(libs.dagger.hilt.compiler)
 
     // Other dependencies
-    testImplementation("junit:junit:4.13.2")
-    androidTestImplementation("androidx.test.ext:junit:1.1.5")
-    androidTestImplementation("androidx.test.espresso:espresso-core:3.5.1")
-    androidTestImplementation("androidx.compose.ui:ui-test-junit4:1.4.3")
-    debugImplementation("androidx.compose.ui:ui-tooling:1.4.3")
-    debugImplementation("androidx.compose.ui:ui-test-manifest:1.4.3")
+    testImplementation(libs.junit)
+    androidTestImplementation(libs.androidx.junit)
+    androidTestImplementation(libs.androidx.espresso.core)
+    androidTestImplementation(libs.ui.test.junit4)
+    debugImplementation(libs.ui.tooling)
+    debugImplementation(libs.ui.test.manifest)
 }
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -1,6 +1,6 @@
 // build.gradle.kts (App-level)
 plugins {
-    id("com.android.application")
+    alias(libs.plugins.android.application)
     alias(libs.plugins.jetbrains.kotlin.android)
     id("dagger.hilt.android.plugin")
     alias(libs.plugins.ksp)
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -1,6 +1,9 @@
+// build.gradle.kts (App-level)
 plugins {
-    alias(libs.plugins.android.application)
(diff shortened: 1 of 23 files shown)


app/build.gradle.kts
```diff
@@ -26,41 +27,47 @@ android {
...
 }
 
 dependencies {
+    implementation(libs.androidx.core.splashscreen)
+    implementation("androidx.core:core-ktx:1.10.1")
+    implementation("androidx.lifecycle:lifecycle-runtime-ktx:2.6.1")
+    implementation("androidx.activity:activity-compose:1.9.0")
+    implementation("androidx.compose.ui:ui:1.4.3")
+    implementation("androidx.compose.ui:ui-tooling-preview:1.4.3")
+    implementation("androidx.compose.material3:material3:1.1.1")
```
philipplackner: Have you thought of using version catalogs here? They're the default nowadays in the latest stable Android Studio version. 
I explain them here: https://youtu.be/MWw1jcwPK3Q?si=yH1n8a3fAKCTxWto

app/build.gradle.kts
```diff
@@ -26,41 +27,47 @@ android {
...
-    androidTestImplementation(libs.androidx.espresso.core)
-    androidTestImplementation(platform(libs.androidx.compose.bom))
-    androidTestImplementation(libs.androidx.ui.test.junit4)
-    debugImplementation(libs.androidx.ui.tooling)
-    debugImplementation(libs.androidx.ui.test.manifest)
-}
\ No newline at end of file
+    // Hilt dependencies
+    implementation("com.google.dagger:hilt-android:2.44")
+    kapt("com.google.dagger:hilt-compiler:2.44")
```
philipplackner: I recommend to use KSP (Kotlin Symbol Processing), which is a faster alternative to kapt. Both Dagger and Room support that.

I explain it here: https://www.youtube.com/watch?v=WVhgUGyUnhQ

app/src/main/java/com/example/tasky/MainActivity.kt
```diff
@@ -1,47 +1,42 @@
...
+    companion object {
+        private const val TAG = "MainActivity"
+    }
     override fun onCreate(savedInstanceState: Bundle?) {
+        Log.d(TAG, "onCreate is called")
+        val splashScreen = installSplashScreen()
         super.onCreate(savedInstanceState)
+        splashScreen.setKeepOnScreenCondition {
+            //TODO add logic to check navigation
+            isLoggedInState.value = splashViewModel.checkIfUserIsLoggedIn()
```
philipplackner: `checkIfUserLoggedIn()` sounds more like a function that purely does something rather than a function that has a return value and delivers a result. Could something like `isUserLoggedIn()` be a better pick here which hints more towards returning a boolean?

app/src/main/java/com/example/tasky/common/presentation/SimpleButton.kt
```diff
@@ -0,0 +1,47 @@
...
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+
+@Composable
+fun SimpleButton(
+    onClick: () -> Unit,
+    modifier: Modifier = Modifier,
+    enabled: Boolean = true,
+    buttonName: String
```
philipplackner: The mandatory arguments should always come at the start of a function, otherwise you can't call it without using named parameters (for composables you want to do that anyways, but it's usually a convention)

app/src/main/java/com/example/tasky/common/presentation/TextBox.kt
```diff
@@ -0,0 +1,45 @@
...
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+
+@Composable
+fun TextBox(hintText: String) {
+    var text by remember { mutableStateOf("") }
```
philipplackner, line 22: Keeping such text state in the composable itself is very dangerous, since this won't survive configuration changes like screen rotations. It also won't survive process death (video about that is in the video portal of this program).

You could bypass both by using `rememberSaveable`, but even that I wouldn't recommend, since the more complex the app becomes the more messy the Compose UI code will become. That's why we typically let our ViewModels manage hosting and updating our state. That not only provides a better separation of concerns, but also lets you test this logic easier since you don't need to spin up a UI test for that.

philipplackner, line 22: Keeping such text state in the composable itself is very dangerous, since this won't survive configuration changes like screen rotations. It also won't survive process death (video about that is in the video portal of this program).

You could bypass both by using `rememberSaveable`, but even that I wouldn't recommend, since the more complex the app becomes the more messy the Compose UI code will become. That's why we typically let our ViewModels manage hosting and updating our state. That not only provides a better separation of concerns, but also lets you test this logic easier since you don't need to spin up a UI test for that.

gradle/libs.versions.toml
```diff
@@ -11,6 +13,9 @@ composeBom = "2023.08.00"
 
 [libraries]
 androidx-core-ktx = { group = "androidx.core", name = "core-ktx", version.ref = "coreKtx" }
+androidx-core-splashscreen = { module = "androidx.core:core-splashscreen", version.ref = "coreSplashscreen" }
+hilt-android = { module = "com.google.dagger:hilt-android", version.ref = "hiltAndroid" }
+hilt-android-compiler = { module = "com.google.dagger:hilt-android-compiler", version.ref = "hiltAndroid" }
```
philipplackner: Ah, you are using version catalogs, how comes you didn't use them in Gradle?

philipplackner (approved):

app/src/main/java/com/example/tasky/feature_splash/domain/SplashViewModel.kt
```diff
@@ -0,0 +1,23 @@
...
+    companion object {
+        private val TAG = "SplashViewModel"
+    }
+
+    var isUserLoggedIn = true
+
+    fun checkIfUserIsLoggedIn(): Boolean {
+        Log.d(TAG, "check if user is logged in")
+        return isUserLoggedIn
+    }
```
philipplackner: If `isUserIsLoggedIn` is public, why have this extra function that returns its value? Why not just refer to `viewModel.isUserLoggedIn` from the UI?

Also, you might want to make that a Compose state or StateFlow, so you can listen to its changes in the Compose UI

philipplackner (approved):

PR #2, opened 2024-06-17
Feature/register screen

sdelator:
**### Newly Added:**

- Button is disabled until criteria for name, email, and password are met 
- Retrofit API to authenticate
- Error Alert Dialogs for API failure
- set up Dagger/Hilt
- create nav graph composable; navigate to others screens
- checking if login fields are valid

**### UI Updates**
image
image
image
image


app/build.gradle.kts
```diff
@@ -1,9 +1,23 @@
...
+
+// Load local.properties file
+val localProperties = Properties()
+val localPropertiesFile = rootProject.file("local.properties")
+if (localPropertiesFile.exists()) {
+    localProperties.load(localPropertiesFile.inputStream())
+}
+
+// Retrieve the API key
+val apiKey: String? = localProperties.getProperty("apiKey")
```
philipplackner, line 13: Nowadays, you can also just do
```
val apiKey = gradleLocalProperties(rootProject.projectDir, providers).getProperty("API_KEY")
```

app/src/main/java/com/example/tasky/MainActivity.kt
```diff
@@ -32,11 +44,40 @@ class MainActivity : ComponentActivity() {
         }
         enableEdgeToEdge()
         setContent {
+            val navController = rememberNavController()
+
+            val startDestination: Any = if (isLoggedInState.value) {
+                LoginNav
+            } else RegisterNav
```
philipplackner, line 51: This doesn't seem correct 🤔 If you're logged in, you should get to agenda

app/src/main/java/com/example/tasky/MainActivity.kt
```diff
@@ -32,11 +44,40 @@ class MainActivity : ComponentActivity() {
...
+                    composable<CalendarNav> {
+                        // todo create Composable
+                        Box(
+                            modifier = Modifier
+                                .fillMaxSize()
+                                .background(Color.Green)
+                        )
+                    }
+                    // todo add other screens
+                }
```
philipplackner, line 70: I like to work with one navigation graph per feature, since then you can conveniently pop the whole graph from the back stack when logging in for example without risking there are some screens left on the back stack.

Works just like defining routes:
```
@Serializable
data object AuthGraph

NavHost(...) {
    navigation<AuthGraph> {
        composable<LoginRoute> {
        
        }
        ...
    }
}
```

app/src/main/java/com/example/tasky/common/domain/TextValidatorExtension.kt
```diff
@@ -0,0 +1,17 @@
+package com.example.tasky.common.domain
+
+import android.util.Patterns
+
+
+fun String.isValidPassword(): Boolean {
+    val passwordPattern = "^(?=.*[a-z])(?=.*[A-Z])(?=.*\\d).{9,}$"
+    return this.matches(passwordPattern.toRegex())
+}
```
philipplackner, line 9: For simple patterns like a valid password, I'm always a friend of doing this manually since it lets you quickly know what makes a valid password. Something like this: 
```
val hasLowerCase = password.any { it.isLowerCase() }
val hasUpperCase = password.any { it.isUpperCase() }
val hasDigit = password.any { it.hasDigit() }
val hasValidLength = password.length in min..max

return hasLowerCase && hasUpperCase && hasDigit && hasValidLength
```
That way, probably even my grandma could tell you what the requirements for a valid password are :D 
Regex makes sense if you can't easily do this manually, but has the cost of making code very unreadable.

app/src/main/java/com/example/tasky/common/domain/TextValidatorExtension.kt
```diff
@@ -0,0 +1,17 @@
...
+
+fun String.isValidEmail(): Boolean {
+    return Patterns.EMAIL_ADDRESS.matcher(this).matches()
+}
```
philipplackner, line 13: Here, you're using an Android-specific dependency (as you can see in the imports) which doesn't really belong into domain. Instead, I'd recommend to create an abstraction for this to keep this class locally unit testable.

Every time, you have implementation details in domain and want to get rid of that, the procedure is the same:

1. You create an abstraction (interface) in domain
2. You write an implementation of that in data (in this case using Patterns.EMAIL_ADDRESS...)
3. You pass the abstraction to the constructor of your class needing it in domain
4. Now you can decide which implementation you pass when you create that class/use case
For example:

*In domain*
```
interface EmailPatternValidator {
    fun isValidEmailPattern(email: String): Boolean
}
```

*In data*
```
class EmailPatternValidatorImpl: EmailPatternValidator {

    override fun isValidEmailPattern(email: String): Boolean {
        return Patterns.EMAIL_ADDRESS.matcher(email).matches()
    }
}
```

Then inject an `EmailPatternValidator` where you want to do the validation.

app/src/main/java/com/example/tasky/common/model/ResolvableResource.kt
```diff
@@ -0,0 +1,22 @@
...
+    }
+}
+
+data class ResolvableStringRes(
+    @StringRes val stringRes: Int
+) : ResolvableResource {
+    override fun resolve(context: Context): String {
+        return context.getString(stringRes)
+    }
+}
```
philipplackner, line 22: Could it make sense to make `ResolvableResource` a `sealed interface` to make it exhaustive when checking for its type?

app/src/main/java/com/example/tasky/common/model/Resource.kt
```diff
@@ -0,0 +1,10 @@
+package com.example.tasky.common.model
+
+sealed class Resource<out T> {
+    data class Success<out T>(val data: T) : Resource<T>()
+    data class Failure<out T>(
+        val errorMessage: String? = "",
+        val throwable: Throwable? = null
+    ) : Resource<T>()
+
+}
```
philipplackner, line 10: For a generic error handling wrapper, I love this version:
```
sealed interface Result<out D, out E: Error> {
    data class Success<out D>(val data: D): Result<D, Nothing>
    data class Error<out E: com.plcoding.tasky.core.domain.util.Error>(val error: E): Result<Nothing, E>
}

inline fun <T, E: Error, R> Result<T, E>.map(map: (T) -> R): Result<R, E> {
    return when(this) {
        is Result.Error -> Result.Error(error)
        is Result.Success -> Result.Success(map(data))
    }
}

fun <T, E: Error> Result<T, E>.asEmptyDataResult(): EmptyResult<E> {
    return map {  }
}

inline fun <T, E: Error> Result<T, E>.onSuccess(action: (T) -> Unit): Result<T, E> {
    return when(this) {
        is Result.Error -> this
        is Result.Success -> {
            action(data)
            this
        }
    }
}
inline fun <T, E: Error> Result<T, E>.onError(action: (E) -> Unit): Result<T, E> {
    return when(this) {
        is Result.Error -> {
            action(error)
            this
        }
        is Result.Success -> this
    }
}

typealias EmptyResult<E> = Result<Unit, E>
```
It handles data and errors in a type-safe way. You'll need a marker interface to mark error enums:
```
interface Error
```
And then you can extend this across the codebase with different error types, e.g.:
```
sealed interface DataError: Error {
    enum class Remote: DataError {
        NOT_FOUND,
        NO_INTERNET,
        UNAUTHORIZED,
        ...
    }
    
    enum class Local: DataError {
        DISK_FULL,
        ...
    }
}
```
Works great:
```
repository.signUp(payload).onSuccess { responseData ->

}.onError { error -> 
    // toUiText() can be an extension of DataError which maps an error to the corresponding message
    _state.update { it.copy(error = error.toUiText()) }
}
```

app/src/main/java/com/example/tasky/common/model/ResponseHandler.kt
```diff
@@ -0,0 +1,33 @@
+package com.example.tasky.common.model
```
philipplackner, line 1: With the chosen type of architecture (presentation, domain, data), it's quite unclear what a class in common->model belongs to. Every class can be clearly attributed to one layer, so I'd also put it in the right layer since otherwise you introduce kind of an arbitrary layer like here where no clear rules exist how interaction with it works.

This class handles responses and therefore clearly relates to data. So the proper package would be common->data->util for example (or similar)

app/src/main/java/com/example/tasky/common/presentation/ErrorAlertDialog.kt
```diff
@@ -0,0 +1,43 @@
...
+import androidx.compose.runtime.MutableState
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.tooling.preview.Preview
+import com.example.tasky.R
+
+@Composable
+fun CreateErrorAlertDialog(
+    showDialog: MutableState<Boolean>,
```
philipplackner, line 16: I'd avoid passing Compose states down to Composables like this, but rather only the raw values (just a Boolean here). You won't find any composable from the official lib that will do this, since you give away all the control you have over the state as don't know in how far it can be changed by sub-composables. 

The logic to show/hide a dialog should therefore be hoisted to the screen composable/screen ViewModel to have this logic at a central place which also makes it easier to test.

app/src/main/java/com/example/tasky/common/presentation/TextBox.kt
```diff
@@ -14,32 +16,76 @@ import androidx.compose.runtime.remember
...
-    var text by remember { mutableStateOf("") }
+fun TextBox(
+    hintText: String,
+    text: String,
+    onValueChange: (String) -> Unit,
+    validator: (String) -> Boolean,
+    isPasswordField: Boolean = false
+) {
+    var isValid by remember { mutableStateOf(validator(text)) }
+    var passwordVisible by remember { mutableStateOf(false) }
```
philipplackner, line 37: Same here, such logic should be put at a central place in your ViewModel. Not only will that make it safe against configuration changes like screen rotations (states with `remember` will be reset after a config change), but you'll also have clear responsibilities for both the pure UI and the logic that decides how the UI should look like. 

As mentioned earlier, that will make it easier to test the UI logic, since for testing a ViewModel you don't need to spin up a real Android device and simulate UI, but rather call some simple functions.

In general, UI should always stay as dumb as possible. So rather than letting the UI decide what a valid password/email is, the ViewModel should decide that and just expose a state like `isEmailValid` which the UI can then decide what that means for the screen.

app/src/main/java/com/example/tasky/common/presentation/TextBox.kt
```diff
@@ -14,32 +16,76 @@ import androidx.compose.runtime.remember
...
+            } else if (isPasswordField) {
+                val image = if (passwordVisible) {
+                    painterResource(id = R.drawable.ic_password_show)
+                } else {
+                    painterResource(id = R.drawable.ic_password_hide)
+                }
+                IconButton(onClick = { passwordVisible = !passwordVisible }) {
+                    Icon(
+                        painter = image,
+                        contentDescription = if (passwordVisible) "Hide password" else "Show password",
```
philipplackner, line 70: If you're using string resources all over the place, it also makes sense to do that for content descriptions, since someone who would rely on a screen reader (for whom content descriptions would become important) would also expect them to be in their local language.

app/src/main/java/com/example/tasky/feature_login/data/remote/BaseHeaderInterceptor.kt
```diff
@@ -0,0 +1,18 @@
+package com.example.tasky.feature_login.data.remote
+
+import com.example.tasky.BuildConfig
+import okhttp3.Interceptor
+import okhttp3.Response
+
+class BaseHeaderInterceptor : Interceptor {
```
philipplackner, line 7: If I gave you this `BaseHeaderInterceptor` class name from another project, could you tell me what this is for? 😄 

Probably not, since if a class contains `Base` it almost certainly breaks the single-responsibility principle OR just has an unclear name. If there is a single responsibility, that could be communicated in the class name. In your case, there is, so calling it `ApiKeyInterceptor` would be much clearer and you wouldn't need to open this to know what's inside.

app/src/main/java/com/example/tasky/feature_login/data/repository/UserRemoteRemoteRepositoryImpl.kt
```diff
@@ -0,0 +1,63 @@
...
+            if (response.isSuccessful) {
+                responseHandler.handleSuccess(Unit)
+            } else {
+                val errorBody = response.errorBody()?.string()
+                val errorMessage =
+                    errorBody?.let { JSONObject(it).getString("message") } ?: "Unknown error"
+                responseHandler.handleFailure(HttpException(response), errorMessage)
+            }
+        } catch (e: Exception) {
+            responseHandler.handleFailure(e, e.message ?: "Unknown error")
```
philipplackner, line 38: It seems like your Retrofit request is returning a `Response` object. When doing so, Retrofit will handle exceptions for you and parse them into this final `Response` object. Therefore, your try/catch isn't necessary here. If however you return the API response directly without wrapping it into this object, you'll need to handle the response yourself.

app/src/main/java/com/example/tasky/feature_login/data/repository/UserRemoteRemoteRepositoryImpl.kt
```diff
@@ -0,0 +1,63 @@
...
+                    responseHandler.handleSuccess(loginUserResponse)
+                } else {
+                    responseHandler.handleFailure(HttpException(response), "Response body is null")
+                }
+            } else {
+                val errorBody = response.errorBody()?.string()
+                val errorMessage =
+                    errorBody?.let { JSONObject(it).getString("message") } ?: "Unknown error"
+                responseHandler.handleFailure(HttpException(response), errorMessage)
+            }
```
philipplackner, line 57: Also, this error logic seems very repetitive. Could it make sense extracting that into a separate extension function or so?

app/src/main/java/com/example/tasky/feature_login/domain/model/AuthenticationInfo.kt
```diff
@@ -0,0 +1,30 @@
...
+    val password: String
+)
+
+data class LoginUserResponse(
+    val accessToken: String,
+    val refreshToken: String,
+    val fullName: String,
+    val userId: String,
+    val accessTokenExpirationTimestamp: Long
+)
```
philipplackner, line 20: Here, you're coupling implementation details with your domain layer which is not allowed. Any DTOs (data transfer objects, like a response model) or DB models must go in your app's data layer since they contain information about the JSON or DB schema. 

The whole goal with domain is that changes in data won't cause changes in domain. However, by putting data models in domain this will be violated, since a change in data (like a change in the JSON schema) would require a change in domain because the fields would need to be renamed.

To fix that, we typically work with mapper functions that take the data model and map it into a domain representation:
```
// In domain
data class LoginUser(
    val accessToken: String,
    val refreshToken: String,
    val fullName: String,
    val userId: String,
    val accessTokenExpirationTimestamp: Long
)

// In data
data class LoginUserResponse(
    val accessToken: String,
    val refreshToken: String,
    val fullName: String,
    val userId: String,
    val accessTokenExpirationTimestamp: Long
)

// In data (LoginMappers.kt)
fun LoginUserResponse.toLoginUser(): LoginUser {
    return LoginUser(...)
}

// In data (AuthRepositoryImpl.kt)
fun login(email: String, password: String): Result<LoginUser> { // <- Returns domain model
    // ...
    return Result.Success(loginUserResponse.toLoginUser()) // <- mapping happens in data layer
}
``` 
For classes like this, this step really feels useless since the domain model looks just like the data model. The important difference is that if there was a change in the JSON schema, just the data model would change, but not the domain model.

In most cases, domain and data models are different though, so don't let this example make you think doing this is always just boilerplate code. Very often, response models work with very simple data which is easy to serialize (imagine a Long timestamp to represent a date - easy to serialize, hard to work with). Mappers can parse such types to data types that are much easier to work with, like `ZonedDateTime`. That makes it very convenient to work with domain models across the app.

app/src/main/java/com/example/tasky/feature_login/domain/repository/UserRemoteRepository.kt
```diff
@@ -0,0 +1,11 @@
+package com.example.tasky.feature_login.domain.repository
+
+import com.example.tasky.common.model.Resource
+import com.example.tasky.feature_login.domain.model.LoginUserInfo
+import com.example.tasky.feature_login.domain.model.LoginUserResponse
+import com.example.tasky.feature_login.domain.model.RegisterUserInfo
+
+interface UserRemoteRepository {
+    suspend fun postRegisterCall(registerUserInfo: RegisterUserInfo): Resource<Unit>
+    suspend fun postLoginCall(loginUserInfo: LoginUserInfo): Resource<LoginUserResponse>
```
philipplackner, line 10: Strictly speaking, the names of these functions also contain implementation details (how the request to the API needs to look like) which domain doesn't need to know. You want to abstract all implementation details out to the data layer to only expose what's really needed (`"Hey, this function is for registering a user, all I need is an email and a password"`)

app/src/main/java/com/example/tasky/feature_login/presentation/LoginScreen.kt
```diff
@@ -12,37 +12,70 @@ import androidx.compose.foundation.layout.padding
...
+import com.example.tasky.feature_login.domain.model.LoginUserInfo
 
 
 @Composable
-@Preview
-fun LoginScreenContent() {
+fun LoginScreenContent(navController: NavController) {
+    // todo figure out why text fields are not saving w/orientation change
+    val loginViewModel: LoginViewModel = hiltViewModel()
+    val authenticationViewModel: AuthenticationViewModel = hiltViewModel()
```
philipplackner, line 49: I'd avoid passing the ViewModel directly to screen composables. This has 2 major reasons:
1. You're making it harder to have a preview, since you'll need to provide a fake ViewModel instance and fakes for everything the ViewModel needs in its constructor.
2. UI tests are also a bit harder since you'll need to provide these fakes.

What I suggest instead:
```
// In your ViewModel
@HiltViewModel
class SampleViewModel @Inject constructor(): ViewModel() {
    private val _state = MutableStateFlow(SampleState()) // Or Compose state alternatively
    val state = _state.asStateFlow()

    fun onEvent(event: SampleEvent) {
        when(event) {
            is SampleEvent.Event1 -> // Handle event1
            is SampleEvent.Event2 -> // Handle event2
            // ...
        }
    }

    // ...
}

// In your NavHost
composable(Routes.sample) {
    SampleScreenRoot(navController)
}

// Your screen composable
@Composable
fun SampleScreenRoot(
    navController: NavHostController,
    viewModel: SampleViewModel = hiltViewModel()
) {
    val state by viewModel.state.collectAsStateWithLifecycle()
    SampleScreen(
        state = state,
        onEvent = viewModel::onEvent
    )
}

@Composable
fun SampleScreen(
    state: SampleState,
    onEvent: (SampleEvent) -> Unit
) {
    // ...
    Button(onClick = { onEvent(SampleEvent.ClickedButtonX) } {
        Text("Click me")
    }
}
```
This way, it's easy to preview and test `SampleScreen` since it doesn't depend on a ViewModel reference.

app/src/main/java/com/example/tasky/feature_login/presentation/LoginScreen.kt
```diff
@@ -54,16 +87,81 @@ fun LoginScreenContent() {
...
+                    onClick = {
+                        // clear focus hides the keyboard
+                        focusManager.clearFocus()
+
+                        authenticationViewModel.logInClicked(
+                            LoginUserInfo(
+                                email = email,
+                                password = password
+                            )
+                        )
```
philipplackner, line 116: The ViewModel already has all the information about the current email and password, so you could save this parameter here and construct this object in the VM itself. Makes the UI more dumb again 😄

app/src/main/java/com/example/tasky/feature_login/presentation/LoginViewModel.kt
```diff
@@ -0,0 +1,42 @@
...
+    val password: StateFlow<String> get() = _password
+
+    private val _isNameValid = MutableStateFlow(false)
+    val isNameValid: StateFlow<Boolean> get() = _isNameValid
+
+    private val _isEmailValid = MutableStateFlow(false)
+    val isEmailValid: StateFlow<Boolean> get() = _isEmailValid
+
+    private val _isPasswordValid = MutableStateFlow(false)
+    val isPasswordValid: StateFlow<Boolean> get() = _isPasswordValid
```
philipplackner, line 26: Your approach of creating a separate StateFlow for everything has a major downside: It's very error-prone since you always need to remember to also update the isValid states when the underlying text field changes. This becomes really dangerous when state changes from many different places. To fix this, you can work with reactivity which can always be applied when you have two states which always change together (whenever email changes, isEmailValid also needs to be recalculated):
```
val isEmailValid = _email.map { email ->
        email.isValidEmail() // <- Each email emission is mapped to this boolean when it changes
    }.stateIn(viewModelScope, SharingStarted.WhileSubscribed(5000L), false)
```

philipplackner (approved):

PR #3, opened 2024-06-22
Feature/register screen - fixes for previous PR suggestions 

sdelator:
- fixed navigation graph
- validator methods to be readable
- api error handling
- made composables dumb


app/src/main/java/com/example/tasky/MainActivity.kt
```diff
@@ -47,37 +57,61 @@ class MainActivity : ComponentActivity() {
...
 }
 
+@Composable
+inline fun <reified T : ViewModel> NavBackStackEntry.sharedViewModel(navController: NavController): T {
+    val navGraphRoute = destination.parent?.route ?: return viewModel()
+    val parentEntry = remember(this) {
+        navController.getBackStackEntry(navGraphRoute)
+    }
+    return viewModel(parentEntry)
+}
```
philipplackner, line 96: What do you need a shared ViewModel here for?

app/src/main/java/com/example/tasky/common/data/util/APIUtilFunctions.kt
```diff
@@ -0,0 +1,34 @@
+package com.example.tasky.common.data.util
+
+import com.google.gson.Gson
+import retrofit2.HttpException
+import javax.inject.Inject
+import javax.inject.Singleton
+
+
+@Singleton
+open class APIUtilFunctions @Inject constructor() {
```
philipplackner, line 10: Why does this class need to be `open`?

app/src/main/java/com/example/tasky/common/data/util/APIUtilFunctions.kt
```diff
@@ -0,0 +1,34 @@
...
+    fun parseErrorResponse(errorBody: String?): String {
+        return if (errorBody != null) {
+            try {
+                val apiErrorResponse = Gson().fromJson(errorBody, ApiErrorResponse::class.java)
+                apiErrorResponse.message
+            } catch (e: Exception) {
+                "$e"
+            }
```
philipplackner, line 18: In general, you don't really need to parse the error body from the response to show it to the user. I would rather use the response HTTP code to determine which error to show. Using the error body won't allow you to localize your error messages, since you always get it as English text back. 
Usually, it's very clear based on the HTTP status what went wrong (or what to show to the user at least). Some examples:
/register + 409 -> Email already exists
/login + 401 -> Invalid credentials
413 -> Uploaded file was too large
500..599 -> Server side error, just display a generic error message (`Oops, something went wrong, please try again`)

app/src/main/java/com/example/tasky/common/data/util/DataError.kt
```diff
@@ -0,0 +1,24 @@
...
+    sealed class Network(val code: Int, private val message: String = "") : DataError {
+        fun formatErrorMessage(): String {
+            return "${this.code}: ${this.message}"
+        }
+
+        class RequestTimeout(message: String = "Request Timeout") : Network(408, message)
+        class Unauthorized(message: String = "Unauthorized") : Network(401, message)
+        class PayloadTooLarge(message: String = "Payload Too Large") : Network(413, message)
+        class Unknown(message: String = "Unknown Error") : Network(-2, message)
+        class NoInternet(message: String = "No Internet") : Network(0, message)
```
philipplackner, line 13: In the data and domain layers, you're generally fine with just knowing the type of error that happened without giving it a specific message. Deciding which error type results in which displayed error message is responsibility of the presentation layer, since that's the only place where an error message is needed, so you could just have an extension function there that parses it:
```
// In presentation
fun DataError.toErrorMessage(): String {
    // ...
}
```

app/src/main/java/com/example/tasky/feature_login/presentation/LoginScreen.kt
```diff
@@ -28,52 +27,50 @@ import androidx.compose.ui.res.stringResource
...
+    val uiState by loginViewModel.uiState.collectAsStateWithLifecycle()
+    val viewState by loginViewModel.viewState.collectAsStateWithLifecycle()
 
-    val viewState by authenticationViewModel.viewState.collectAsState()
     val email by loginViewModel.email.collectAsState()
     val password by loginViewModel.password.collectAsState()
 
     val isEmailValid by loginViewModel.isEmailValid.collectAsState()
     val isPasswordValid by loginViewModel.isPasswordValid.collectAsState()
+    val isPasswordVisible by loginViewModel.isPasswordVisible.collectAsState()
```
philipplackner, line 55: I'd recommend to use `collectAsStateWithLifecycle()` that since it's the safest way to collect flows in Android. For that, you'll need this dependency:
```
implementation "androidx.lifecycle:lifecycle-runtime-compose:2.x.x"
```
Then use it like this:
```
val state by viewModel.state.collectAsStateWithLifecycle()
```
If you have a simple flow it shouldn't make a difference, but as soon as you convert a normal `Flow` into a `StateFlow` with `stateIn`, the Flow operators will be executed even when the app is in the background (which you typically don't want when dealing with UI updates).

app/src/main/java/com/example/tasky/feature_login/presentation/LoginViewModel.kt
```diff
@@ -1,12 +1,51 @@
...
+import android.util.Log
 import androidx.lifecycle.ViewModel
-import com.example.tasky.common.domain.isValidEmail
+import androidx.lifecycle.viewModelScope
+import com.example.tasky.common.data.EmailPatternValidatorImpl
+import com.example.tasky.common.data.util.Result
 import com.example.tasky.common.domain.isValidName
 import com.example.tasky.common.domain.isValidPassword
+import com.example.tasky.feature_login.data.model.LoginUserInfo
+import com.example.tasky.feature_login.data.model.RegisterUserInfo
```
philipplackner, line 11: Here, you're coupling your data layer to presentation which is not allowed with this type of architecture. The direction of dependencies is this:
Presentation -> Domain <- Data

There's no arrow from presentation to data, so you should also have no imports of data in presentation (or vice versa). Instead of accessing this login and register info here, you can just pass the required strings to the data layer in this case. If it's a more complex class with many fields, you would create a domain model for this which you are allowed to access here.

app/src/main/java/com/example/tasky/feature_login/presentation/LoginViewModel.kt
```diff
@@ -16,27 +55,100 @@ class LoginViewModel : ViewModel() {
...
+                    // emit a viewState to show ErrorMessage
+                    val errorMsg = result.error.formatErrorMessage()
+                    _viewState.emit(AuthenticationViewState.Failure(errorMsg))
+                }
+            }
+        }
+    }
+
+    fun onShowErrorDialog(message: String) {
+        _uiState.value = _uiState.value.copy(showErrorDialog = true, dialogMessage = message)
```
philipplackner, line 148: In general, I'd always use `_stateFlow.update { it.copy(...) }` to avoid race conditions. This is how it could happen:

*StateFlow holds value A*
- Thread 1 calls `copy()` on StateFlow with value A
- Before Thread 1 assigns its changed copy, Thread 2 updates StateFlow with value B
- Now, Thread 1 assigns its changed copy, which is still based on value A though, but the latest value is already value B. Therefore, value B would never be considered here.

Not always an issue, but I worked on highly stateful apps with frequently changing states, in which this really bit me, since such bugs are super hard to find.
Using `_state.value = _state.value.copy(...)` is fine if you always call this from a single thread (which you do here). But in ViewModels, you're very likely to not do that as soon as you'd change the state inside a `viewModelScope`.

philipplackner (approved):

app/src/main/java/com/example/tasky/common/data/util/APIUtilFunctions.kt
sdelator: very good point and I did not think about this: **"Using the error body won't allow you to localize your error messages, since you always get it as English text back."**

PR #4, opened 2024-06-22
Feature/register screen - fixes for previous PR suggestions part 2

sdelator:
- fix error handling
- fix wrong coupling of Presentation -> Domain <- Data

### DIFF
--- a/app/src/main/java/com/example/tasky/feature_login/data/repository/UserRemoteRemoteRepositoryImpl.kt
+++ b/app/src/main/java/com/example/tasky/feature_login/data/repository/UserRemoteRemoteRepositoryImpl.kt
@@ -24,9 +24,14 @@ class UserRemoteRemoteRepositoryImpl(
         println("hello from the repository. The appname is $appName")
     }
 
-    override suspend fun registerUser(registerUserInfo: RegisterUserInfo): Result<Unit, DataError.Network> {
+    override suspend fun registerUser(
+        name: String,
+        email: String,
+        password: String
+    ): Result<Unit, DataError.Network> {
         return try {
-            val response = api.registerUser(registerUserInfo = registerUserInfo)
+            val response =
+                api.registerUser(registerUserInfo = RegisterUserInfo(name, email, password))
             if (response.isSuccessful) {
                 Result.Success(Unit)
             } else {
@@ -46,9 +51,12 @@ class UserRemoteRemoteRepositoryImpl(
         }
     }
 
-    override suspend fun logInUser(loginUserInfo: LoginUserInfo): Result<LoginUserResponse, DataError.Network> {
+    override suspend fun logInUser(
+        email: String,
+        password: String
+    ): Result<LoginUserResponse, DataError.Network> {
         return try {
-            val response = api.loginUser(loginUserInfo = loginUserInfo)
+            val response = api.loginUser(loginUserInfo = LoginUserInfo(email, password))
             val loginUserResponse = response.body()
             if (response.isSuccessful && loginUserResponse != null) {
                 Result.Success(loginUserResponse)
--- a/app/src/main/java/com/example/tasky/feature_login/data/repository/UserRemoteRemoteRepositoryImpl.kt
+++ b/app/src/main/java/com/example/tasky/feature_login/data/repository/UserRemoteRemoteRepositoryImpl.kt
@@ -11,7 +11,6 @@ import com.example.tasky.feature_login.data.model.RegisterUserInfo
 import com.example.tasky.feature_login.data.remote.TaskyApi
 import com.example.tasky.feature_login.domain.repository.UserRemoteRepository
 import okio.IOException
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/example/tasky/feature_login/domain/model/AuthenticationViewState.kt
```diff
@@ -1,7 +1,9 @@
 package com.example.tasky.feature_login.domain.model
```
sdelator, line 1: is this coupling fine? domain and data

app/src/main/java/com/example/tasky/feature_login/data/repository/UserRemoteRemoteRepositoryImpl.kt
```diff
@@ -24,54 +23,50 @@ class UserRemoteRemoteRepositoryImpl(
...
+                    413 -> DataError.Network.PAYLOAD_TOO_LARGE
+                    else -> DataError.Network.UNKNOWN
                 }
-
                 Result.Error(error)
             }
-        } catch (e: HttpException) {
-            val error = apiUtilFunctions.getHttpErrorMessage(e)
-            Result.Error(error)
         } catch (e: IOException) {
```
sdelator, line 45: do I still need this catch? this part is unclear. maybe I can remove the whole try/catch block and just keep the if statement

philipplackner: I think you do still need this, since it's not directly related to a response result, but about being able to make the request in the first place.. But I'm not 100% sure myself here, you can easily try this out by executing the request with airplane mode turned on. If you get the no internet error, you need this catch.

app/src/main/java/com/example/tasky/feature_login/domain/model/AuthenticationViewState.kt
philipplackner: No, the direction is like this:
presentation -> domain <- data

The goal with this is to let changes in frequently changing layers (data and presentation) have no impact on less frequently changing layers (domain). If domain depends on data however, a change in your data layer could also cause a change in your domain layer which we want to prevent.

In your case, this class is misplaced in domain however. As the name `AuthenticationViewState` already indirectly says, this is related to something presentational. The only purpose is to use this to show certain data to the user. Any UI, ViewModels, UI state, UI events/actions belong in the presentation layer.

app/src/main/java/com/example/tasky/feature_login/domain/model/AuthenticationViewState.kt
```diff
@@ -1,7 +1,9 @@
...
 
+import com.example.tasky.common.data.util.DataError
+
 open class AuthenticationViewState {
     object Success : AuthenticationViewState()
-    data class Failure(val message: String) : AuthenticationViewState()
+    data class Failure(val dataError: DataError) : AuthenticationViewState()
     object Loading : AuthenticationViewState()
```
philipplackner, line 8: Also if you have a class with a fixed set of options, better make that a sealed class/interface. Then you get proper smart casting and when-exhaustiveness :)

app/src/main/java/com/example/tasky/feature_login/presentation/LoginScreen.kt
```diff
@@ -136,7 +138,10 @@ fun LoginScreenContent(
 
             is AuthenticationViewState.Failure -> {
                 // Show an Alert Dialog with API failure Error code/message
-                val message = (viewState as AuthenticationViewState.Failure).message
+                val message =
+                    (viewState as AuthenticationViewState.Failure).dataError.toLoginErrorMessage(
+                        context = LocalContext.current
+                    )
```
philipplackner, line 144: With the sealed class recommendation from above, this cast would not be necessary here, since you made the type check before 👍

philipplackner (approved):

app/src/main/java/com/example/tasky/feature_login/domain/model/AuthenticationViewState.kt
sdelator: this is hard concept for me to grasp.. I'm rechecking all files and making sure they are in the correct location. Will resolve this in next PR.

app/src/main/java/com/example/tasky/feature_login/data/repository/UserRemoteRemoteRepositoryImpl.kt
sdelator: I tried with wifi off and it needs the catch IO Exception block

app/src/main/java/com/example/tasky/feature_login/presentation/LoginScreen.kt
sdelator: ohhh nice!

PR #7, opened 2024-06-24
Add navigation login to register

sdelator:
### **Newly Added**

- navigation from login to register

### **UI Changes**
image


app/src/main/java/com/example/tasky/feature_login/data/repository/UserRemoteRepositoryImpl.kt
```diff
@@ -37,6 +37,7 @@ class UserRemoteRemoteRepositoryImpl(
                 val error = when (response.code()) {
                     401 -> DataError.Network.UNAUTHORIZED
                     408 -> DataError.Network.REQUEST_TIMEOUT
+                    409 -> DataError.Network.CONFLICT
                     413 -> DataError.Network.PAYLOAD_TOO_LARGE
                     else -> DataError.Network.UNKNOWN
                 }
```
philipplackner, line 43: Could it make sense to create a separate extension function for such response code parsing? It'll be the same for every request in the end :)

philipplackner (approved):

PR #8, opened 2024-06-25
Agenda/logout/profileIcon

sdelator:
### **Newly Added:**
- Agenda Screen skeleton 
- Agenda Toolbar with a temporary Text (month) and a Profile Icon
- Profile Icon is Clickable 
- added logic to call remote logout 
- added viewstate for type of response (success, failure)
- if success navigate to auth graph (login)
- if failure display error dialog 

### **UI Updates**
image
image


app/src/main/java/com/example/tasky/MainActivity.kt
```diff
@@ -90,9 +77,20 @@ fun NavGraphBuilder.authGraph(navController: NavController, loginViewModel: Logi
     }
 }
 
+fun NavGraphBuilder.calendarGraph(navController: NavController, agendaViewModel: AgendaViewModel) {
+    composable<AgendaNav> {
+        AgendaContent(navController = navController, agendaViewModel = agendaViewModel)
+    }
+}
```
philipplackner, line 84: Extracting this into a separate function doesn't make it a navigation graph yet 😄 
You'll need this inside:
```
navigation<CalendarGraph>(startDestination = AgendaNav) {
    composable<AgendaNav> {
    
    }
}
```

app/src/main/java/com/example/tasky/feature_agenda/presentation/Agenda.kt
```diff
@@ -0,0 +1,115 @@
...
+import com.example.tasky.R
+import com.example.tasky.common.domain.error.DataError
+import com.example.tasky.common.presentation.CreateErrorAlertDialog
+import com.example.tasky.common.presentation.LoadingSpinner
+
+@Composable
+fun AgendaContent(
+    navController: NavController,
+    agendaViewModel: AgendaViewModel = hiltViewModel()
+) {
```
philipplackner, line 37: I think I've commented on that before, but you're still coupling your nav controller and ViewModel to the screen composable which will not make it previewable. You can quickly check this by adding a `@Preview` composable, it won't work. 
Instead, better create a root composable which accepts these arguments that break the preview and pass down arguments that don't:
```
@Composable
fun AgendaRoot(
    navController: NavController,
    viewModel: AgendaViewModel = hiltViewModel()
) {
    // Initialize state, etc.
    
    AgendaContent(
        state = state,
        onAction = onAction
    )
}

@Composable
fun AgendaContent(
    state: AgendaState,
    onAction: (AgendaAction) -> Unit
) {
    // Actual UI
}
```
`AgendaAction` can be a sealed interface with all different user actions that could happen:
```
sealed interface AgendaAction {
    data object OnLogoutClick: AgendaAction
    data class OnCreateAgendaItemClick(val itemType: AgendaItemType): AgendaAction
    ...
}
```
After that, you can preview `AgendaContent` without issues, since it's not bound to a ViewModel or NavController anymore.

app/src/main/java/com/example/tasky/feature_agenda/presentation/Agenda.kt
```diff
@@ -0,0 +1,115 @@
...
+    val uiState by agendaViewModel.uiState.collectAsStateWithLifecycle()
+    val viewState by agendaViewModel.viewState.collectAsStateWithLifecycle()
```
philipplackner, line 39: What's the difference between view and UI state? Sounds very misleading to me

app/src/main/java/com/example/tasky/feature_agenda/presentation/Agenda.kt
```diff
@@ -0,0 +1,115 @@
...
+        when (viewState) {
+            is AgendaViewState.Loading -> {
+                // Show a loading indicator
+                LoadingSpinner()
+            }
+
+            is AgendaViewState.Success -> {
+                // Handle success by navigating to LoginScreen
+                navController.navigate(AuthNavRoute)
+            }
```
philipplackner, line 83: Here you have a side-effect which has to be handled properly with `LaunchedEffect`. Otherwise, this code might execute multiple times when the state remains success and a recomposition happens.

Also it's not clear to me why you are logging out here. The code doesn't communicate that at all that the success is all about logging out 🤔

app/src/main/java/com/example/tasky/feature_agenda/presentation/Agenda.kt
```diff
@@ -0,0 +1,115 @@
...
+            is AgendaViewState.Failure -> {
+                // Show an Alert Dialog with API failure Error code/message
+                val message =
+                    (viewState as AgendaViewState.Failure).dataError.toLogOutErrorMessage(
+                        context = LocalContext.current
+                    )
+                LaunchedEffect(message) {
+                    agendaViewModel.onShowErrorDialog(message)
+                }
+            }
```
philipplackner, line 94: Interesting concept, how do you distinguish between different types of success? What speaks against integrating this in your UI state so you don't need multiple states per screen? You could just make this extra fields like there.

philipplackner (approved):

app/src/main/java/com/example/tasky/feature_agenda/presentation/Agenda.kt
sdelator: looking back at this, I'm not sure why I made those states in ViewStates. ViewStates should reflect specific UI states. I went back and changed some states. Yes we only need one UI state and can use extra fields. what was i thinking..

app/src/main/java/com/example/tasky/feature_agenda/presentation/Agenda.kt
sdelator: I'm not sure why I put 2 separate. Will merge these

app/src/main/java/com/example/tasky/feature_agenda/presentation/Agenda.kt
sdelator: I think I finally understand this concept. I will make improvements in next PR!

PR #9, opened 2024-06-28
Feature/logout/ - Fixes for previous PR suggestions

sdelator:

- Fix decoupling my Composables was coupling nav controller and ViewModel
- added Previews to ensure it works
- fix viewStates to make sense 
- separate viewmodels for login and register. It was causing composable states to copy onto each screen

### DIFF
--- a/app/src/main/java/com/example/tasky/feature_agenda/presentation/Agenda.kt
+++ b/app/src/main/java/com/example/tasky/feature_agenda/presentation/Agenda.kt
@@ -14,8 +14,8 @@ import androidx.compose.material3.Card
 import androidx.compose.material3.CardDefaults
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
-import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
+import androidx.compose.runtime.livedata.observeAsState
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.graphics.Color
@@ -25,7 +25,6 @@ import androidx.compose.ui.unit.dp
 import androidx.hilt.navigation.compose.hiltViewModel
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
 import androidx.navigation.NavController
-import com.example.tasky.AuthNavRoute
 import com.example.tasky.R
 import com.example.tasky.common.domain.error.DataError
 import com.example.tasky.common.presentation.CreateErrorAlertDialog
@@ -37,6 +36,9 @@ fun AgendaRoot(
     agendaViewModel: AgendaViewModel = hiltViewModel()
 ) {
     val viewState by agendaViewModel.viewState.collectAsStateWithLifecycle()
+    val viewEvent by agendaViewModel.viewEvent.observeAsState(null)
+
+    val showDialog by agendaViewModel.showDialog.collectAsStateWithLifecycle()
 
     AgendaContent(
         getInitials = { agendaViewModel.getInitials() },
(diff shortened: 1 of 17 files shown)


app/src/main/java/com/example/tasky/feature_agenda/presentation/Agenda.kt
```diff
@@ -31,22 +32,64 @@ import com.example.tasky.common.presentation.CreateErrorAlertDialog
...
+        null -> println("no action")
+    }
+
+    LaunchedEffect(Unit) {
+        agendaViewModel.viewEvent.collect { event ->
+            if (event is AgendaViewEvent.NavigateToLoginScreen) {
+                navController.navigate(AuthNavRoute)
+            }
+        }
+    }
```
philipplackner, line 76: Strictly speaking, you need something like this to collect one-time events 100% safely without risking to lose some:
```
@Composable
private fun <T> ObserveAsEvents(flow: Flow<T>, onEvent: (T) -> Unit) {
    val lifecycleOwner = LocalLifecycleOwner.current
    LaunchedEffect(flow, lifecycleOwner.lifecycle) {
        lifecycleOwner.repeatOnLifecycle(Lifecycle.State.STARTED) {
            withContext(Dispatchers.Main.immediate) {
                flow.collect(onEvent)
            }
        }
    }
}
```
For more info, I talk about the problem in this video: https://youtu.be/njchj9d_Lf8?si=0to4KwRHzEIECID8

app/src/main/java/com/example/tasky/feature_login/presentation/LoginScreen.kt
```diff
@@ -63,8 +62,78 @@ fun LoginScreenContent(
...
+
+        null -> println("no action")
+    }
+
+    LaunchedEffect(Unit) {
+        loginViewModel.viewEvent.collect { event ->
+            when (event) {
+                is LoginViewEvent.NavigateToAgenda -> {
+                    navController.navigate(CalendarNavRoute)
+                }
```
philipplackner, line 113: Here it's important to also clear the auth graph from the back stack, since otherwise the user could get back to the login screen after being logged in. You can do so with 
```
navController.navigate(CalendarNavRoute) {
    setPopUpTo(AuthGraph) { // or whatever your graph is called like
        inclusive = true
    }
}
```

app/src/main/java/com/example/tasky/feature_login/presentation/LoginViewModel.kt
```diff
@@ -68,9 +52,12 @@ class LoginViewModel @Inject constructor(
...
 
-    fun onNameChange(newName: String) {
-        _name.value = newName
-    }
+    private val _showDialog = MutableStateFlow(false)
+    val showDialog: StateFlow<Boolean> get() = _showDialog
+
+    // viewEvent triggered by API response
+    private val _viewEvent = MutableSharedFlow<LoginViewEvent>()
+    val viewEvent: SharedFlow<LoginViewEvent> = _viewEvent
```
philipplackner, line 60: For UI events, you might be better off using a `Channel` instead of a `SharedFlow`. Channels include an integrated buffer, so if an event can't be collected (e.g. because the app is in the background), the channel keeps it in the buffer to send when the collector is back. So if the app is in the background and you want to show an error toast, the user would see that when going back. With SharedFlows that event would simply be lost, unless you set something to the `replay` parameter in the shared flow's constructor.

With channels it would look like this:
```
private val _viewEvent = Channel<LoginViewEvent>()
val viewEvent = _viewEvent.receiveAsFlow()
```

philipplackner (approved):

PR #11, opened 2024-06-28
Creating dataStore

sdelator:
**### Newly Added:**

- Create DataStore
- store userName in dataStore and format it for ProfileIcon

### DIFF
--- a/app/src/main/java/com/example/tasky/common/data/TaskyStateImpl.kt
+++ b/app/src/main/java/com/example/tasky/common/data/TaskyStateImpl.kt
@@ -21,5 +21,13 @@ class TaskyStateImpl @Inject constructor(
         return appPreferences.get(PreferencesKeys.REFRESH_TOKEN).first() ?: ""
     }
 
+    override suspend fun setName(name: String) {
+        appPreferences.put(PreferencesKeys.NAME, name)
+    }
+
+    override suspend fun getName(): String {
+        return appPreferences.get(PreferencesKeys.NAME).first() ?: ""
+    }
+
 
 }
--- a/app/src/main/java/com/example/tasky/common/data/TaskyStateImpl.kt
+++ b/app/src/main/java/com/example/tasky/common/data/TaskyStateImpl.kt
@@ -1,0 +1,25 @@
+package com.example.tasky.common.data
+
+import com.example.tasky.common.domain.TaskyState
+import kotlinx.coroutines.flow.first
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class TaskyStateImpl @Inject constructor(
+    private val appPreferences: PreferenceHelper,
+) : TaskyState {
+    companion object {
+        private const val TAG = "TaskyState"
+    }
+
+    override suspend fun setRefreshToken(refreshToken: String) {
+        appPreferences.put(PreferencesKeys.REFRESH_TOKEN, refreshToken)
+    }
+
+    override suspend fun getRefreshToken(): String {
+        return appPreferences.get(PreferencesKeys.REFRESH_TOKEN).first() ?: ""
+    }
+
+
+}
--- a/app/src/main/java/com/example/tasky/feature_agenda/presentation/AgendaViewModel.kt
+++ b/app/src/main/java/com/example/tasky/feature_agenda/presentation/AgendaViewModel.kt
@@ -4,6 +4,7 @@ import android.util.Log
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import com.example.tasky.common.domain.Result
+import com.example.tasky.common.domain.TaskyState
 import com.example.tasky.feature_login.domain.repository.UserRemoteRepository
 import dagger.hilt.android.lifecycle.HiltViewModel
 import kotlinx.coroutines.channels.Channel
@@ -15,7 +16,8 @@ import javax.inject.Inject
 
 @HiltViewModel
 class AgendaViewModel @Inject constructor(
-    private val userRemoteRepository: UserRemoteRepository
+    private val userRemoteRepository: UserRemoteRepository,
+    private val taskyState: TaskyState
 ) : ViewModel() {
 
     companion object {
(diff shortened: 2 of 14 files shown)


app/src/main/java/com/example/tasky/common/data/TaskyStateImpl.kt
```diff
@@ -0,0 +1,33 @@
+package com.example.tasky.common.data
+
+import com.example.tasky.common.domain.TaskyState
+import kotlinx.coroutines.flow.first
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class TaskyStateImpl @Inject constructor(
```
philipplackner, line 9: Not the most intuitive name IMO 😄 What do you think about `SessionStateManager`?

app/src/main/java/com/example/tasky/feature_agenda/presentation/AgendaViewModel.kt
```diff
@@ -59,9 +68,24 @@ class AgendaViewModel @Inject constructor(
...
+                initials = "${name[0]}${name[1]}"
+            } else {
+                for (i in nameSplit.indices) {
+                    if (i == 0 || i == nameSplit.lastIndex) {
+                        initials += nameSplit[i].first()
+                    }
+                }
+            }
+            _initials.value = initials.uppercase()
+        }
```
philipplackner, line 88: Could it make sense extracting that logic into a separate function independent of the ViewModel? Then you could reuse it for the agenda detail screen as well (attendee section). 
Also, your code will break for a name with leading, trailing or double spaces in between.

app/src/main/java/com/example/tasky/feature_agenda/presentation/AgendaViewModel.kt
```diff
@@ -59,9 +68,24 @@ class AgendaViewModel @Inject constructor(
...
+
+            if (nameSplit.size == 1) {
+                val name = nameSplit.first()
+                initials = "${name[0]}${name[1]}"
+            } else {
+                for (i in nameSplit.indices) {
+                    if (i == 0 || i == nameSplit.lastIndex) {
+                        initials += nameSplit[i].first()
+                    }
+                }
```
philipplackner, line 85: You don't need a loop here if you know you only need to access the first and last word. This would also do it:
```
initials += "${nameSplit[0][0]}${nameSplit[nameSplit.lastIndex][0]}"
```

app/src/test/java/com/example/tasky/feature_agenda/presentation/AgendaViewModelTest.kt
```diff
@@ -0,0 +1,83 @@
...
+    @RelaxedMockK
+    private lateinit var taskyState: TaskyState
+
+    @RelaxedMockK
+    private lateinit var userRemoteRepository: UserRemoteRepository
+
+    private lateinit var viewModel: AgendaViewModel
+
+    private val testDispatcher = StandardTestDispatcher()
+    private val testScope = TestScope(testDispatcher)
```
philipplackner, line 28: All objects in a test should be initialized in setUp. You don't know if they contain internal state, but if they do, the state changes of one test could break another test if not reinitialized.

app/src/test/java/com/example/tasky/feature_agenda/presentation/AgendaViewModelTest.kt
```diff
@@ -0,0 +1,83 @@
...
+
+    @OptIn(ExperimentalCoroutinesApi::class)
+    @Before
+    fun setUp() {
+        MockKAnnotations.init(this, relaxUnitFun = true)
+        Dispatchers.setMain(testDispatcher)
```
philipplackner, line 34: Similarly for this, you should undo this after each test case again (`Dispatchers.resetMain()`)

app/src/test/java/com/example/tasky/feature_agenda/presentation/AgendaViewModelTest.kt
```diff
@@ -0,0 +1,83 @@
...
+import kotlinx.coroutines.test.setMain
+import org.junit.Before
+import org.junit.Test
+
+class AgendaViewModelTest {
+    @RelaxedMockK
+    private lateinit var taskyState: TaskyState
+
+    @RelaxedMockK
+    private lateinit var userRemoteRepository: UserRemoteRepository
```
philipplackner, line 23: I personally got away from using mocks as much as possible. They're good in two scenarios:
1. You're testing legacy code which isn't written with testability in mind.
2. You're testing code you don't own (e.g. you need to make a library's function return something specific and it's not extendable)

In all other cases, I found fakes/stubs much more reliable and less error prone. I can't count how often my tests broke at some point, not because there was a bug, but because the mock itself broke (for example when the mocked functionality changed/was extended, there were unmocked function calls, etc.). 
Having a fake instead will make you implement new functionality before you can even run your tests and also doesn't rely on that much 3rd party library magic as it's just a plain Kotlin class.

app/src/test/java/com/example/tasky/feature_agenda/presentation/AgendaViewModelTest.kt
```diff
@@ -0,0 +1,83 @@
...
+        coEvery { taskyState.getName() } returns "firstName"
+
+        // when
+        runTest {
+            viewModel.getInitials()
+        }
+        val initials = viewModel.initials.value
+
+        // then
+        assertEquals("FI", initials)
```
philipplackner, line 51: I would recommend to look into assertion libraries, since the default assertions aren't so readable for more complex ones. My personal favorite is AssertK. You can add it like this:
1. Version catalog declaration
```
[versions]
...
assertk = "0.28.1"

[libraries]
...
assertk = { module = "com.willowtreeapps.assertk:assertk", version.ref = "assertk" }
```
2. Reference it in your Gradle file:
```
dependencies {
    ...
    testImplementation(libs.assertk)
    androidTestImplementation(libs.assertk)
}
```
3. Then your assertions could look like this for example:
```
assertThat(initials).isEqualTo("PL")

assertThat(myList).hasSize(5)

assertThat(myList).contains(element)

assertThat(myBoolean).isTrue()
```
Reads like a book :D

philipplackner (approved):

PR #12, opened 2024-06-29
Creating datastore - Fixes for previous PR suggestions PR suggestions Fixes

sdelator:
- Fix unit test
- move function to get initials into a common place

### DIFF
--- a/app/src/test/java/com/example/tasky/fakes/FakeSessionStateManager.kt
+++ b/app/src/test/java/com/example/tasky/fakes/FakeSessionStateManager.kt
@@ -1,0 +1,27 @@
+package com.example.tasky.fakes
+
+import com.example.tasky.common.domain.SessionStateManager
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.first
+
+class FakeSessionStateManager : SessionStateManager {
+    private val refreshTokenFlow = MutableStateFlow<String?>(null)
+    private val nameFlow = MutableStateFlow<String?>(null)
+
+    override suspend fun setRefreshToken(refreshToken: String) {
+        refreshTokenFlow.value = refreshToken
+    }
+
+    override suspend fun getRefreshToken(): String {
+        return refreshTokenFlow.first() ?: ""
+    }
+
+    override suspend fun setName(name: String) {
+        nameFlow.value = name
+    }
+
+    override suspend fun getName(): String {
+        return nameFlow.first() ?: ""
+    }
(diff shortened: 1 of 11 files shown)


app/src/test/java/com/example/tasky/fakes/FakeSessionStateManager.kt
```diff
@@ -0,0 +1,27 @@
...
+    private val refreshTokenFlow = MutableStateFlow<String?>(null)
+    private val nameFlow = MutableStateFlow<String?>(null)
+
+    override suspend fun setRefreshToken(refreshToken: String) {
+        refreshTokenFlow.value = refreshToken
+    }
+
+    override suspend fun getRefreshToken(): String {
+        return refreshTokenFlow.first() ?: ""
+    }
```
philipplackner, line 17: Instead of making non-existent strings empty, I'd make them `null` instead. To me it's not obvious that just because a token is empty there is no token (by design I would think there always is one). Don't be afraid to use nulls what they are intended to be used for (indicating that a value doesn't exist).

Also, you can just access a state flow's value with `refreshTokenFlow.value` without suspending.

app/src/test/java/com/example/tasky/feature_agenda/presentation/AgendaViewModelTest.kt
```diff
@@ -1,83 +1,86 @@
...
-        viewModel =
-            AgendaViewModel(userRemoteRepository = userRemoteRepository, taskyState = taskyState)
-    }
+        val testDispatcher = StandardTestDispatcher()
 
-    @Test
-    fun formatInitials_given_firstName() {
-        // given
-        coEvery { taskyState.getName() } returns "firstName"
+        Dispatchers.setMain(testDispatcher)
```
philipplackner, line 30: You can extract that into a `MainCoroutineRule` which you can apply for test classes that always calls this for you (together with resetMain()):
```
@OptIn(ExperimentalCoroutinesApi::class)
class MainCoroutineRule(
    val testDispatcher: TestDispatcher = StandardTestDispatcher()
) : TestWatcher() {

    override fun starting(description: Description?) {
        Dispatchers.setMain(testDispatcher)
    }

    override fun finished(description: Description?) {
        Dispatchers.resetMain()
    }
}
```
Then apply it in your test like this:
```
class MyTest {

    @get:Rule
    val mainCoroutineRule = MainCoroutineRule()
}
```

philipplackner (approved):

PR #14, opened 2024-06-30
Logout/drop down menu

sdelator:
**### Newly Added:**
- Separate unauthenticated and authenticated API calls because one uses accessToken interceptor 
- Create drop-down menu composable

### **UI Updates:**
image


app/src/main/java/com/example/tasky/common/data/SessionStateManagerManagerImpl.kt
```diff
@@ -13,6 +15,16 @@ class SessionStateManagerManagerImpl @Inject constructor(
...
     }
 
+    override suspend fun setAccessToken(accessToken: String) {
+        appPreferences.put(PreferencesKeys.ACCESS_TOKEN, accessToken)
+    }
+
+    override fun getAccessToken(): String? {
+        return runBlocking {
+            appPreferences.get(PreferencesKeys.ACCESS_TOKEN).firstOrNull()
+        }
```
philipplackner, line 25: Avoid using runBlocking here when you have no certainty that this is not called from the UI thread. Instead just expose a suspend function like above and execute this on a "real" scope (usually viewModelScope). Or even better just expose the flow directly so you can also observe the value.

philipplackner (approved):

PR #17, opened 2024-07-05
Agenda/horizontal calendar

sdelator:
### Newly Added
- Horizontal Scrollable Calendar
- Date gets highlighted on Click 
- If user selects a date on the month picker on the toolbar, then the horizontal scrollbar updates to the selected date 
- utilize viewState to reduce number of flows per variable

### **UI Changes**
image

### DIFF
--- a/app/src/main/java/com/example/tasky/common/data/CalendarHelperImpl.kt
+++ b/app/src/main/java/com/example/tasky/common/data/CalendarHelperImpl.kt
@@ -1,0 +1,28 @@
+package com.example.tasky.common.data
+
+import com.example.tasky.common.domain.CalendarHelper
+import com.example.tasky.feature_agenda.presentation.model.CalendarDay
+import java.time.LocalDate
+import java.util.Locale
+import javax.inject.Inject
+import javax.inject.Singleton
+
+@Singleton
+class CalendarHelperImpl @Inject constructor() : CalendarHelper {
+    override fun getCalendarDaysForMonth(year: Int, month: Int): List<CalendarDay> {
+        val firstDayOfMonth = LocalDate.of(year, month, 1)
+        val daysInMonth = firstDayOfMonth.lengthOfMonth()
+
+        val calendarDays = mutableListOf<CalendarDay>()
+        for (date in 1..daysInMonth) {
+            val currentDate = firstDayOfMonth.withDayOfMonth(date)
+            val dayOfWeek = currentDate.dayOfWeek.getDisplayName(
+                java.time.format.TextStyle.NARROW,
+                Locale.getDefault()
+            )
+            calendarDays.add(CalendarDay(dayOfWeek, date))
+        }
+
+        return calendarDays
+    }
+}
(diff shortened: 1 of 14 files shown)


app/src/main/java/com/example/tasky/common/data/CalendarHelperImpl.kt
```diff
@@ -0,0 +1,28 @@
...
+            val dayOfWeek = currentDate.dayOfWeek.getDisplayName(
+                java.time.format.TextStyle.NARROW,
+                Locale.getDefault()
+            )
+            calendarDays.add(CalendarDay(dayOfWeek, date))
+        }
+
+        return calendarDays
+    }
+}
```
philipplackner, line 28: What made you feel like this belongs in the data layer? I would argue that date formatting is something needed only for the presentation layer to display something, so could that be the better place? You also don't need the abstraction in that case.

app/src/main/java/com/example/tasky/feature_agenda/presentation/AgendaViewModel.kt
```diff
@@ -14,99 +15,96 @@ import kotlinx.coroutines.flow.MutableStateFlow
...
-    val initials: StateFlow<String> get() = _initials
-
-    private val _monthSelected = MutableStateFlow<String>(getCurrentMonth())
-    val monthSelected: StateFlow<String> get() = _monthSelected
-
-    private val _dateDialogState = MutableStateFlow<MaterialDialogState>(MaterialDialogState())
-    val dateDialogState: StateFlow<MaterialDialogState> get() = _dateDialogState
+    // not displayed on screen; only necessary for business logic
+    private val _yearSelected = MutableStateFlow(LocalDate.now().year)
+    val yearSelected: StateFlow<Int> get() = _yearSelected
```
philipplackner, line 43: If it doesn't need to be accessed elsewhere, you can also leave it a single private property

app/src/main/java/com/example/tasky/feature_agenda/presentation/AgendaViewModel.kt
```diff
@@ -14,99 +15,96 @@ import kotlinx.coroutines.flow.MutableStateFlow
...
                 is Result.Error -> {
                     println("failed logout :(")
-                    // emit a viewState to show ErrorMessage
-                    _showDialog.value = true
-                    _viewState.emit(AgendaViewState.ErrorDialog(result.error))
+                    _viewState.value = _viewState.value.copy(
+                        showLoadingSpinner = false,
+                        showErrorDialog = true,
+                        dataError = result.error
+                    )
```
philipplackner, line 76: In general, I'd always use `_stateFlow.update { it.copy(...) }` to avoid race conditions. This is how it could happen:

*StateFlow holds value A*
- Thread 1 calls `copy()` on StateFlow with value A
- Before Thread 1 assigns its changed copy, Thread 2 updates StateFlow with value B
- Now, Thread 1 assigns its changed copy, which is still based on value A though, but the latest value is already value B. Therefore, value B would never be considered here.

Not always an issue, but I worked on highly stateful apps with frequently changing states, in which this really bit me, since such bugs are super hard to find.
Using `_state.value = _state.value.copy(...)` is fine if you always call this from a single thread (which you do here). But in ViewModels, you're very likely to not do that as soon as you'd change the state inside a `viewModelScope`.

philipplackner (approved):

PR #18, opened 2024-07-07
Implement Refresh Token Session ~ Expiration Timestamp

sdelator:
### Newly Added
- add logic for token refresh according to expiration timestamp
- added temporary button to check authentication. If access token expires, then refresh the token

### **UI Changes**
image


app/src/main/java/com/example/tasky/common/data/remote/AuthInterceptor.kt
```diff
@@ -1,20 +1,26 @@
...
+    private val tokenManager: TokenManager
 ) : Interceptor {
     companion object {
         const val AUTH = "Authorization"
     }
 
     override fun intercept(chain: Interceptor.Chain): Response {
+        runBlocking {
+            tokenManager.shouldRefreshToken()
+        }
```
philipplackner: This code looks like it doesn't do anything.. From a function `shouldRefreshToken()` I would expect that it returns whether to refresh the token, but not actually refreshing it. That would be a side-effect since the function doesn't communicate that.
If you want this function to actually refresh the token if necessary, why not call it like that? Something like `refreshIfNecessary` will be very clear :)

app/src/main/java/com/example/tasky/common/domain/TokenManager.kt
```diff
@@ -0,0 +1,46 @@
...
+
+    suspend fun shouldRefreshToken() {
+        val now = ZonedDateTime.now()
+        val currentTimeInMilliseconds = now.toInstant().toEpochMilli()
+        val expirationTime = sessionStateManager.getAccessTokenExpiration().first() ?: 0L
+
+        if (currentTimeInMilliseconds > expirationTime) {
+            refreshToken()
+        }
+    }
```
philipplackner: Alternatively, you can just make this return `currentTimeInMilliseconds > expirationTime` and then have something like this:
```
if(shouldRefreshToken()) {
    refreshToken()
}
```
Is just as readable as what I shared above

philipplackner (approved):

PR #19, opened 2024-07-07
Horizontal calendar - Fixes for previous PR suggestions 

sdelator:
- fix misuse of abstraction; use object instead
- only use year in viewModel. Thats where I use it anyways
- use _stateFlow.update { it.copy(...) } to avoid race conditions

### DIFF
--- a/app/src/main/java/com/example/tasky/feature_agenda/presentation/AgendaViewModel.kt
+++ b/app/src/main/java/com/example/tasky/feature_agenda/presentation/AgendaViewModel.kt
@@ -3,9 +3,9 @@ package com.example.tasky.feature_agenda.presentation
 import android.util.Log
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
-import com.example.tasky.common.domain.CalendarHelper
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/example/tasky/feature_agenda/presentation/AgendaViewModel.kt
```diff
@@ -69,42 +69,55 @@ class AgendaViewModel @Inject constructor(
...
 
     fun updateDateDialogState(dialogState: MaterialDialogState) {
-        _viewState.value = _viewState.value.copy(datePickerDialogState = dialogState)
+        _viewState.update {
+            it.copy(datePickerDialogState = dialogState)
+        }
     }
 
-    fun updateDateSelected(date: LocalDate) {
+    fun updateDateSelected(month: Int, day: Int, year: Int? = null) {
```
philipplackner, line 102: I would recommend to get into using the Java LocalDateTime API which makes working with dates a lot easier :) 

You would then just pass a `dateTime: LocalDateTime` to this function from which you can derive all those values and do date math.

philipplackner (approved):

PR #21, opened 2024-07-11
Agenda/Floating Action Button 

sdelator:
**### Newly Added**
- create FAB composable and add a dropdown menu
- add navigation element for event, task, and reminder 

### **UI Changes**
image
image


app/src/main/java/com/example/tasky/feature_agenda/presentation/AgendaViewModel.kt
```diff
@@ -165,4 +172,22 @@ class AgendaViewModel @Inject constructor(
...
+                Action.Task -> {
+                    _viewEvent.send(AgendaViewEvent.NavigateToTaskScreen)
+                }
+
+                Action.Reminder -> {
+                    _viewEvent.send(AgendaViewEvent.NavigateToReminderScreen)
+                }
+            }
+        }
+    }
```
philipplackner, line 192: You already have access to the nav controller in the UI, why take this detour over the ViewModel just to send an event back?

philipplackner (approved):

PR #23, opened 2024-07-15
Event Screen/UI

sdelator:
### **### Newly Added**

- UI Only for Event screen
- Set toolbar date title based on date selection from agenda screen

### **### UI Changes**
image

### DIFF
--- a/app/src/main/java/com/example/tasky/event/presentation/Event.kt
+++ b/app/src/main/java/com/example/tasky/event/presentation/Event.kt
@@ -6,6 +6,7 @@ import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.safeDrawingPadding
 import androidx.compose.foundation.layout.size
@@ -14,6 +15,7 @@ import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.foundation.verticalScroll
 import androidx.compose.material3.Card
 import androidx.compose.material3.CardDefaults
+import androidx.compose.material3.Divider
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Alignment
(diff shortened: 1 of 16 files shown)


app/src/main/java/com/example/tasky/event/presentation/Event.kt
```diff
@@ -0,0 +1,152 @@
...
+    navController: NavController,
+    date: Long,
+    eventViewModel: EventViewModel = hiltViewModel()
+) {
+//    val viewState by eventViewModel.viewState.collectAsStateWithLifecycle()
+    EventContent(
+        dateOnToolbar = date.convertMillisToDate(),
+        onToolbarAction = {}
+    )
+}
```
philipplackner, line 49: One of the tougher decisions to make for this project is whether to use one screen per agenda item type or one screen for all agenda items. 2 years ago, I was also team single screen per type, but my mind has changed :D

My concern was that having a single screen makes it harder to read and update, but it actually got a lot easier:
If you look at the UI hierarchy of the items, reminder and task screens are almost identical while the event screen also has the same elements with a few extra composables. 

The complete root screen structure is no more than this:
```
HeaderSection()
TitleSection()
DescriptionSection()
if(item is Event) {
    PhotoSection()
}
StartDateSection()
if(item is Event) {
    EndDateSection()
}
ReminderSection()
if(item is Event) {
    AttendeeSection()
}
DeleteButton()
```
As long as you extract the UI into single components, it remains very readable. Same goes for the underlying ViewModel, you will save so much code by just having one ViewModel, since the logic of what all items have in common only needs to be written once. And if you'd extend that screen with another type of agenda item, that would also be easier and wouldn't require you to duplicate most of this to another screen.

app/src/main/java/com/example/tasky/event/presentation/EventViewState.kt
```diff
@@ -0,0 +1,21 @@
...
+    val startTime: String = "",
+    val endTime: String = "",
+    val startDate: String = "",
+    val endDate: String = "",
+    val reminder: Int = 0,
+    val eventViewMode: EventViewMode = EventViewMode.EDITABLE,
+    val filterSelected: Filter = Filter.ALL
+)
+
+enum class Filter(type: String) {
```
philipplackner, line 15: I'd probably also specify what kind of filter this is (like `AttendeeFilter`)

philipplackner (approved):

PR #24, opened 2024-07-16
Event/UI - Fixes for previous PR suggestions

sdelator:
- convert Event screen into a reusable ActionScreen for event, task, and reminders

### DIFF
--- a/app/src/main/java/com/example/tasky/actions/presentation/ActionRoot.kt
+++ b/app/src/main/java/com/example/tasky/actions/presentation/ActionRoot.kt
@@ -1,4 +1,4 @@
-package com.example.tasky.event.presentation
+package com.example.tasky.actions.presentation
 
 import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Box
@@ -16,6 +16,7 @@ import androidx.compose.material3.Card
 import androidx.compose.material3.CardDefaults
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.getValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.clip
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/example/tasky/actions/presentation/ActionRoot.kt
```diff
@@ -26,40 +27,45 @@ import androidx.compose.ui.tooling.preview.Preview
...
 import com.example.tasky.common.presentation.DateLineItem
 import com.example.tasky.common.presentation.GrayDivider
 import com.example.tasky.common.presentation.RightCarrotIcon
-import com.example.tasky.event.presentation.components.EmptyPhotos
+import com.example.tasky.common.presentation.TitleSection
+import com.example.tasky.common.presentation.model.Action
 
 @Composable
-fun EventRoot(
+fun ActionRoot(
```
philipplackner, line 42: What does ActionRoot mean? Why call this action? 🤔 

We're talking about agenda details here after all, right?

app/src/main/java/com/example/tasky/actions/presentation/ActionViewState.kt
```diff
@@ -0,0 +1,21 @@
...
+    val attendeeFilterSelected: AttendeeFilter = AttendeeFilter.ALL
+)
+
+enum class AttendeeFilter(type: String) {
+    ALL("All"), GOING("Going"), NOT_GOING("Not Going")
+}
+
+enum class ActionViewMode {
+    NON_EDITABLE, EDITABLE, ATTENDEE, ATTENDEE_EDITABLE
+}
```
philipplackner, line 21: There are actually two distinct states hidden here which are not mutually exclusive (you can be both an attendee and it's editable). You've encoded that with two extra enums, but an enum is not the right data structure in that case, since it models mutually exclusive things by default.
If something is not mutually exclusive, better work with booleans:
```
val isEditable: Boolean = false
val isLocalUserAttendee: Boolean = false
```

app/src/main/java/com/example/tasky/actions/presentation/ActionRoot.kt
sdelator: yeah! I can rename

PR #25, opened 2024-07-17
Action/time picker

sdelator:
**### Newly Added**
- Create the TimePicker Component and add to onClick()
- add date picker to onClick()

**### UI Changes**


app/src/main/java/com/example/tasky/common/presentation/TimePicker.kt
```diff
@@ -0,0 +1,103 @@
...
+        }
+    ) {
+        timepicker(
+            colors = customTimePickerColors()
+        ) {
+            dialogState.hide()
+        }
+    }
+    /* This should be called in an onClick or an Effect */
+    dialogState.show()
```
philipplackner, line 83: So why don't you do it then? :D

app/src/main/java/com/example/tasky/event/presentation/EventViewState.kt
```diff
@@ -0,0 +1,27 @@
...
+    val filterSelected: Filter = Filter.ALL
+)
+
+enum class Filter(type: String) {
+    ALL("All"), GOING("Going"), NOT_GOING("Not Going")
+}
+
+enum class EventViewMode {
+    NON_EDITABLE, EDITABLE, ATTENDEE, ATTENDEE_EDITABLE
+}
```
philipplackner, line 27: Is this a duplication of the ActionViewMode from the last PR? If so, why?

philipplackner (approved):

app/src/main/java/com/example/tasky/event/presentation/EventViewState.kt
sdelator: change is fixed now in commit

PR #26, opened 2024-07-18
Agenda details/Text-Editing screen

sdelator:
### **### Newly Added**
- added a text editor screen 
- save and back button functionality

### DIFF
--- a/app/src/main/java/com/example/tasky/common/presentation/editing/Editing.kt
+++ b/app/src/main/java/com/example/tasky/common/presentation/editing/Editing.kt
@@ -32,22 +32,24 @@ import com.example.tasky.R
 import com.example.tasky.common.domain.Constants
 import com.example.tasky.common.presentation.GrayDivider
 import com.example.tasky.common.presentation.LeftCarrotIcon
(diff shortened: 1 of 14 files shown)


app/src/main/java/com/example/tasky/common/presentation/editing/Editing.kt
```diff
@@ -0,0 +1,207 @@
...
+        text = stringResource(R.string.save),
+        color = colorResource(id = R.color.tasky_green),
+        fontWeight = FontWeight.SemiBold,
+        fontSize = 16.sp,
+        modifier = Modifier.clickable { onClick() }
+    )
+}
+
+@Composable
+fun getPlaceholderText(textFieldType: String, agendaDetailsType: String): String {
```
philipplackner: Could it make sense to work with enums here instead to make this less error-prone?

philipplackner (approved):

app/src/main/java/com/example/tasky/common/presentation/editing/Editing.kt
sdelator: note: convert string back to enum after navigation

PR #29, opened 2024-07-25
Event/Photos logic 

sdelator:
**### Newly Added**
- Added on empty photo Click functionality
- opens phone Gallery and user can select up to 10 images at a time
- compresses photo
- checks if compressed photo size > 1MB then send a toast message and skip 

**### UI Changes**


app/src/main/java/com/example/tasky/agenda_details/domain/ImageCompression.kt
```diff
@@ -0,0 +1,23 @@
+package com.example.tasky.agenda_details.domain
+
+import android.content.Context
+import android.graphics.Bitmap
+import android.graphics.drawable.BitmapDrawable
+import android.graphics.drawable.Drawable
+import android.net.Uri
```
philipplackner, line 7: Having lots of Android imports here in domain which is not allowed since it should be isolated 😄 
I'd recommend to leave this up to the data layer. You can work with an abstraction that you put in domain, like
```
interface ImageCompressor {
    suspend fun compress(bytes: ByteArray): ByteArray
}
```

app/src/main/java/com/example/tasky/agenda_details/domain/ImageCompression.kt
```diff
@@ -0,0 +1,23 @@
...
+        val stream = ByteArrayOutputStream()
+        bitmap.compress(Bitmap.CompressFormat.JPEG, quality, stream)
+        return stream.toByteArray()
+    }
+
+    fun uriToDrawable(context: Context, uri: Uri): Drawable? {
+        val contentResolver = context.contentResolver
+        val inputStream = contentResolver.openInputStream(uri)
+        return Drawable.createFromStream(inputStream, uri.toString())
+    }
```
philipplackner, line 22: Also both these are blocking functions. It would make sense to make these suspend functions to avoid accidentally blocking the main thread. You can do so by just adding the suspend keyword and wrapping each function's code into `withContext(Dispatchers.IO)`

app/src/main/java/com/example/tasky/agenda_details/presentation/AgendaDetailsRoot.kt
```diff
@@ -58,7 +65,10 @@ fun AgendaDetailsRoot(
     agendaDetailsViewModel: AgendaDetailsViewModel = hiltViewModel()
 ) {
     val viewState by agendaDetailsViewModel.viewState.collectAsStateWithLifecycle()
+    val photosUri by agendaDetailsViewModel.photosUri.collectAsStateWithLifecycle()
+    val photoSkipCount by agendaDetailsViewModel.photoSkipCount.collectAsStateWithLifecycle()
```
philipplackner, line 69: Why did you decide to not make these part of the screen state?

app/src/main/java/com/example/tasky/agenda_details/presentation/AgendaDetailsRoot.kt
```diff
@@ -70,10 +80,23 @@ fun AgendaDetailsRoot(
             agendaDetailsViewModel.onDateSelected(selectedDate, dialogState, timeType)
         }
 
+    val compressAndAddImage: (Context, List<Uri>) -> Unit =
+        { context, uri ->
+            agendaDetailsViewModel.compressAndAddImage(context, uri)
```
philipplackner, line 85: ViewModels should also not receive references to the context, since then you won't be able to write local unit tests anymore. The moment you have an Android SDK reference (like Context) in a class, you need to test it on a real Android device, build an APK, etc. That's a lot slower than being able to run a test just on the JVM.

Therefore, only include Android references in classes that really need them (typically only the UI and certain classes in data)

app/src/main/java/com/example/tasky/agenda_details/presentation/AgendaDetailsViewModel.kt
```diff
@@ -119,4 +139,37 @@ class AgendaDetailsViewModel @Inject constructor() : ViewModel() {
...
+        viewModelScope.launch {
+            var skipped = 0
+            val newCompressedList = uris.mapNotNull { uri ->
+                val drawable = imageCompressionUseCase.uriToDrawable(context, uri)
+                val originalBitmap = (drawable as BitmapDrawable).bitmap
+                Log.d(TAG, "Original Bitmap Size in bytes: ${originalBitmap.byteCount}")
+
+                val compressedByteArray = withContext(Dispatchers.IO) {
+                    imageCompressionUseCase.compressImage(drawable)
+                }
```
philipplackner, line 157: As in my previous comment, I'd recommend to move this withContext block into the `compressImage` function itself. You added it here, which is good and won't cause any bugs in this context, but it violates the concept of main-safety (a suspend function should be safe to call from the main thread without blocking it). Just because you added the withContext block here doesn't mean that you'll always remember doing that when calling this function from somewhere else. If the withContext block is in the compressImage function however, it doesn't matter from which dispatcher you call it, since it makes sure by itself that it always runs on the right thread. More about that here:
https://www.youtube.com/watch?v=4_fDefmI3yI

app/src/main/java/com/example/tasky/agenda_details/presentation/AgendaDetailsViewModel.kt
```diff
@@ -119,4 +139,37 @@ class AgendaDetailsViewModel @Inject constructor() : ViewModel() {
...
+
+                val compressedBitmap =
+                    BitmapFactory.decodeByteArray(compressedByteArray, 0, compressedByteArray.size)
```
philipplackner, line 160: This is also a blocking call which will block the main thread here, since you're running it on Dispatchers.Main

app/src/main/java/com/example/tasky/agenda_details/presentation/components/Photos.kt
```diff
@@ -39,19 +60,102 @@ fun EmptyPhotos() {
...
+    compressedImages: List<Bitmap?>,
+    compressAndAddImage: (Context, List<Uri>) -> Unit,
+    photoSkipCount: Int,
+    onAddPhotosClick: () -> Unit
+) {
+    val context = LocalContext.current
+
+    LaunchedEffect(selectedPhotoUris) {
+        compressAndAddImage(context, selectedPhotoUris)
+    }
```
philipplackner, line 92: Doesn't feel very efficient to me to always compress **all** images even if you add or remove just one. Why not just compress the one you added and leave the rest unchanged?

Also, could it make sense to move this call into the ViewModel, since this doesn't really need to get triggered from the UI, right? The ViewModel knows, when the Uris change.

philipplackner (approved):

app/src/main/java/com/example/tasky/agenda_details/presentation/AgendaDetailsRoot.kt
sdelator: I am going to fix this, add variables to my viewState that has any value that can change and has an effect on the screen. follow MVI architectural pattern.

app/src/main/java/com/example/tasky/agenda_details/presentation/AgendaDetailsRoot.kt
sdelator: removing context from viewmodel

PR #30, opened 2024-07-26
Photo/pr suggestion - Fixes for previous PR suggestions

sdelator:
- use abstraction for my Image compressor. There was too many imports in domain
- convert compress functions into suspend functions
- fix viewState to hold any value that can change and has an effect on the screen ~ follow MVI architectural pattern
- remove Context from ViewModel
- move compress function call from Composable to ViewModel

### DIFF
--- a/app/src/main/java/com/example/tasky/agenda_details/data/ImageCompressionImpl.kt
+++ b/app/src/main/java/com/example/tasky/agenda_details/data/ImageCompressionImpl.kt
@@ -1,21 +1,23 @@
-package com.example.tasky.agenda_details.domain
+package com.example.tasky.agenda_details.data
 
 import android.content.Context
 import android.graphics.Bitmap
 import android.graphics.drawable.BitmapDrawable
 import android.graphics.drawable.Drawable
 import android.net.Uri
+import com.example.tasky.agenda_details.domain.ImageCompressionUseCase
 import java.io.ByteArrayOutputStream
 
-class ImageCompressionUseCase {
-    fun compressImage(drawable: Drawable, quality: Int = 100): ByteArray {
+class ImageCompressionUseCaseImpl : ImageCompressionUseCase {
+    override fun compressImage(drawable: Drawable): ByteArray {
+        val quality = 100
         val bitmap = (drawable as BitmapDrawable).bitmap
         val stream = ByteArrayOutputStream()
         bitmap.compress(Bitmap.CompressFormat.JPEG, quality, stream)
         return stream.toByteArray()
     }
 
-    fun uriToDrawable(context: Context, uri: Uri): Drawable? {
+    override fun uriToDrawable(context: Context, uri: Uri): Drawable? {
         val contentResolver = context.contentResolver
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/example/tasky/agenda_details/data/ImageCompressionImpl.kt
```diff
@@ -0,0 +1,31 @@
...
+import android.graphics.drawable.BitmapDrawable
+import android.graphics.drawable.Drawable
+import android.net.Uri
+import com.example.tasky.agenda_details.domain.ImageCompressionUseCase
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+import java.io.ByteArrayOutputStream
+
+class ImageCompressionUseCaseImpl(private val contentResolver: ContentResolver) :
+    ImageCompressionUseCase {
```
philipplackner, line 14: While the way you did this is totally correct, a use case is meant for business logic and should be in domain. So instead you'd just need to give this a different name, like `ImageCompressor` and `BitmapCompressor` for the implementation 👍

app/src/main/java/com/example/tasky/agenda_details/data/ImageCompressionImpl.kt
```diff
@@ -0,0 +1,31 @@
...
+    override suspend fun compressImage(drawable: Drawable): ByteArray {
+        return withContext(Dispatchers.IO) {
+            val quality = 100
```
philipplackner, line 17: If you pass a quality of 100%, nothing will be compressed :D

app/src/main/java/com/example/tasky/agenda_details/presentation/AgendaDetailsViewModel.kt
```diff
@@ -141,23 +132,22 @@ class AgendaDetailsViewModel @Inject constructor(
...
 
-                val compressedByteArray = withContext(Dispatchers.IO) {
-                    imageCompressionUseCase.compressImage(drawable)
-                }
+                val compressedByteArray = imageCompressionUseCase.compressImage(drawable)
 
-                val compressedBitmap =
+                val compressedBitmap = withContext(Dispatchers.IO) {
                     BitmapFactory.decodeByteArray(compressedByteArray, 0, compressedByteArray.size)
+                }
```
philipplackner, line 150: It's not just the Context that should stay out of ViewModels, but any class from the Android SDK (like Bitmaps), since these classes can't be used in local unit tests. 

I don't think you need a bitmap here at all, since you could just pass the compressed byte array to the UI which loads it as an image (I think Coil does that out of the box for byte arrays)

philipplackner (approved):

app/src/main/java/com/example/tasky/agenda_details/data/ImageCompressionImpl.kt
sdelator: ah true!

PR #31, opened 2024-07-26
Photo/pr suggestion 2 - Fixes for previous PR suggestions

sdelator:

- rename file because its not a use-case 
- pass quality of 80 for image compression to work better
- pass the compressed byte array to the UI which loads it as an image


app/src/main/java/com/example/tasky/agenda_details/data/ImageCompressorImpl.kt
```diff
@@ -5,16 +5,16 @@ import android.graphics.Bitmap
...
 import java.io.ByteArrayOutputStream
 
-class ImageCompressionUseCaseImpl(private val contentResolver: ContentResolver) :
-    ImageCompressionUseCase {
+class ImageCompressorImpl(private val contentResolver: ContentResolver) :
+    ImageCompressor {
     override suspend fun compressImage(drawable: Drawable): ByteArray {
         return withContext(Dispatchers.IO) {
-            val quality = 100
+            val quality = 80
```
philipplackner, line 17: From an image compressor I'd probably expect that I can pass the quality, since otherwise this is quite inflexible

philipplackner (approved):

app/src/main/java/com/example/tasky/agenda_details/data/ImageCompressorImpl.kt
sdelator: true, I will add this ability 👍

PR #32, opened 2024-07-28
Photo/delete

sdelator:
### **### Newly Added**

- remove all Android imports from ViewModel (use String instead of Uri)
- Create Photo Detail screen 
- add photo detail functionality (delete photo)

### **### UI Changes**
image

image


app/src/main/java/com/example/tasky/agenda_details/data/ImageCompressorImpl.kt
```diff
@@ -12,20 +12,24 @@ import java.io.ByteArrayOutputStream
...
         }
     }
 
-    override suspend fun uriToDrawable(uri: Uri): Drawable? {
+    override suspend fun uriStringToDrawable(uri: String): Drawable {
         return withContext(Dispatchers.IO) {
-            val inputStream = contentResolver.openInputStream(uri)
-            Drawable.createFromStream(inputStream, uri.toString())
+            val inputStream = contentResolver.openInputStream(Uri.parse(uri))
+            requireNotNull(inputStream) { "Failed to open input stream for URI: $uri" }
```
philipplackner, line 27: I'd be very careful with non-null assertions. This will make your app crash of the Uri doesn't exist for some reason. Why not just have a simple null-check? Also I would recommend to use `.use` for the input stream since that closes it automatically after use (you keep it open forever here :D):
```
return contentResolver.openInputStream(Uri.parse(uri))?.use { inputStream -> 
    Drawable.createFromStream(inputStream, uri)
}
```

app/src/main/java/com/example/tasky/agenda_details/presentation/AgendaDetailsViewModel.kt
```diff
@@ -128,35 +126,99 @@ class AgendaDetailsViewModel @Inject constructor(
...
+            // update compressedImageList
+            val oldImageList = _viewState.value.byteArrayImageList
+            val newImageList = oldImageList.toMutableList().apply {
+                removeAt(imageIndex)
+            }
+
+            _viewState.update {
+                it.copy(
+                    uriImageList = newUriList,
+                    byteArrayImageList = newImageList,
```
philipplackner, line 204: Could it maybe also make sense to think of a different data structure that also properly supports remote photos (that have a key and a URL rather than a local Uri?)

app/src/main/java/com/example/tasky/agenda_details/presentation/components/PhotoDetail.kt
```diff
@@ -0,0 +1,157 @@
...
+import androidx.navigation.NavController
+import coil.compose.AsyncImage
+import com.example.tasky.R
+import com.example.tasky.agenda_details.presentation.PhotoDetailAction
+import com.example.tasky.common.domain.Constants
+
+@Composable
+fun PhotoDetailRoot(
+    navController: NavController,
+    image: String
```
philipplackner, line 44: Calling an argument `image` and making it a string makes it very hard to know what this now represents 😄 Is it a file path? A content Uri? A URL? An image byte array encoded as a string? Unless we know, we don't know

philipplackner (approved):

app/src/main/java/com/example/tasky/agenda_details/presentation/components/PhotoDetail.kt
sdelator: good call, will rename!

app/src/main/java/com/example/tasky/agenda_details/data/ImageCompressorImpl.kt
sdelator: added the null check

app/src/main/java/com/example/tasky/agenda_details/presentation/components/PhotoDetail.kt
sdelator: renamed!

PR #33, opened 2024-07-30
use Sealed class for photos - Fixes for previous PR suggestions

sdelator:
- Use a sealed class to load remote and local images


app/src/main/java/com/example/tasky/agenda_details/data/model/EventDetails.kt
```diff
@@ -0,0 +1,20 @@
...
+)
+
+sealed class PhotoType {
+    data class LocalPhoto(
+        val byteArray: ByteArray,
+        val uri: String
+    ) : PhotoType()
+
+    data class RemotePhoto(val key: String, val url: String) : PhotoType()
+}
```
philipplackner, line 20: Looks great 👍 A few suggestions:
1. The class does not just represent a photo type, but a complete photo with everything that belongs to that, so could a name like `Photo` be more appropriate since it contains more than just the type information?
2. If you decide to use `Uri.parse(...)` for both local Uri and URL, you can make the Uri string a shared parameter by putting it into the constructor of the sealed class.

app/src/main/java/com/example/tasky/agenda_details/presentation/AgendaDetailsRoot.kt
```diff
@@ -97,11 +97,7 @@ fun AgendaDetailsRoot(
             agendaDetailsViewModel.onDateSelected(selectedDate, dialogState, timeType)
         }
 
-    val titleText = if (!title.isNullOrEmpty()) title else getDefaultTitle(agendaItemType)
-    val itemDescription =
-        if (!description.isNullOrEmpty()) description else getDefaultDescription(agendaItemType)
-
-    val maxPhotosSelection = maxOf(2, maxPhotoCount - viewState.uriImageList.size)
+    val maxPhotosSelection = maxOf(2, maxPhotoCount - viewState.photos.size)
```
philipplackner, line 100: This one I don't understand. If the user already picked 10 photos, then `maxPhotoCount - viewState.photos.size` will be 0 and the outcome will be that you can pick two more photos.. 🤔

sdelator: I was getting a crash when I set it to 1

java.lang.IllegalArgumentException: Max items must be higher than 1

PR #34, opened 2024-07-31
Event/create api

sdelator:
### **Newly Added**
- api call to create event
- add functionality to Save button on agendaDetails screen


app/src/main/java/com/example/tasky/agenda_details/data/repository/AgendaDetailsRemoteRepositoryImpl.kt
```diff
@@ -0,0 +1,80 @@
...
+    ): Result<EventResponse, DataError.Network> {
+        return try {
+            val gson = Gson()
+            val json = gson.toJson(eventDetails)
+            val requestBody = json.toRequestBody("application/json; charset=utf-8".toMediaType())
+
+            val photoParts = photosByteArray.mapIndexed { index, byteArray ->
+                val requestFile =
+                    byteArray.toRequestBody("image/jpeg".toMediaTypeOrNull(), 0, byteArray.size)
+                MultipartBody.Part.createFormData("photo$index", "image$index.jpg", requestFile)
```
philipplackner, line 34: Do you let the user only pick JPEGs or could this break for PNGs?

philipplackner (approved):

app/src/main/java/com/example/tasky/agenda_details/data/repository/AgendaDetailsRemoteRepositoryImpl.kt
sdelator: I'm adding a condition so Png can be allowed too

PR #35, opened 2024-07-31
Event create api/pr fix  - Fixes for previous PR suggestions 

sdelator:
- include logic for .png images too
- fix my sealed class for Photos
- add condition to allow user to select multiple photos if 1 < x < 10 else select single photo

### DIFF
--- a/app/src/main/java/com/example/tasky/agenda_details/data/model/EventDetails.kt
+++ b/app/src/main/java/com/example/tasky/agenda_details/data/model/EventDetails.kt
@@ -10,11 +10,8 @@ data class EventDetails(
     val attendeeIds: List<String>
 )
 
-sealed class PhotoType {
-    data class LocalPhoto(
-        val byteArray: ByteArray,
-        val uri: String
-    ) : PhotoType()
+sealed class Photo(open val uri: String) {
+    data class LocalPhoto(val byteArray: ByteArray, override val uri: String) : Photo(uri = uri)
 
-    data class RemotePhoto(val key: String, val url: String) : PhotoType()
+    data class RemotePhoto(val key: String, val url: String) : Photo(uri = url)
 }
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/example/tasky/agenda_details/data/model/EventDetails.kt
```diff
@@ -10,11 +10,8 @@ data class EventDetails(
     val attendeeIds: List<String>
 )
 
-sealed class PhotoType {
-    data class LocalPhoto(
-        val byteArray: ByteArray,
-        val uri: String
-    ) : PhotoType()
+sealed class Photo(open val uri: String) {
```
philipplackner, line 13: It seems you put this class in your data layer, how do you want to access this in presentation without violating the architecture? This actually belongs in domain 😄

app/src/main/java/com/example/tasky/agenda_details/data/repository/AgendaDetailsRemoteRepositoryImpl.kt
```diff
@@ -29,9 +29,18 @@ class AgendaDetailsRemoteRepositoryImpl(
             val requestBody = json.toRequestBody("application/json; charset=utf-8".toMediaType())
 
             val photoParts = photosByteArray.mapIndexed { index, byteArray ->
+                val mimeType = when (byteArray.take(3)) {
+                    listOf(0xFF.toByte(), 0xD8.toByte(), 0xFF.toByte()) -> "image/jpeg"
+                    listOf(0x89.toByte(), 0x50.toByte(), 0x4e.toByte()) -> "image/png"
+                    else -> throw Exception("invalid Image format")
+                }
```
philipplackner, line 36: Interesting approach, which may work for the happy path, but what if an image ends with `.jpeg`? That's also a valid abbreviation for JPG files. The "correct" way to extract mime types is with Android's `MimeTypeMap`

app/src/main/java/com/example/tasky/agenda_details/presentation/AgendaDetailsViewModel.kt
```diff
@@ -3,11 +3,10 @@ package com.example.tasky.agenda_details.presentation
 import androidx.lifecycle.ViewModel
 import androidx.lifecycle.viewModelScope
 import com.example.tasky.agenda_details.data.model.EventDetails
-import com.example.tasky.agenda_details.data.model.PhotoType
+import com.example.tasky.agenda_details.data.model.Photo
```
philipplackner, line 6: Here, you can see the architectural violation, since presentation should not have a dependency on data. With this architecture, dependencies always point inwards:
presentation -> domain <- data

sdelator: oh no! good catch, I have to keep an eye out for this

PR #36, opened 2024-08-01
Agenda/api  (load agenda, load Full Agenda)

sdelator:
### Newly Added

- agenda api calls (load agenda, load Full Agenda)
- show cards on screen 
- format date strings for card 
- add Lazy Column for Card

### UI Changes
image

### DIFF
--- a/app/src/main/java/com/example/tasky/agenda_details/presentation/utils/DateTimeHelper.kt
+++ b/app/src/main/java/com/example/tasky/agenda_details/presentation/utils/DateTimeHelper.kt
@@ -36,11 +36,11 @@ object DateTimeHelper {
         return duration.seconds
     }
 
-    fun getEpochSecondsFromDateAndTime(date: String, time: String): Long {
+    fun getEpochMillisecondsFromDateAndTime(date: String, time: String): Long {
         val dateFormatter = DateTimeFormatter.ofPattern("MMM dd yyyy")
         val localTime = LocalTime.parse(time)
         val localDate = LocalDate.parse(date, dateFormatter)
         val zdt = ZonedDateTime.of(localDate, localTime, ZoneOffset.UTC)
-        return zdt.toEpochSecond()
+        return zdt.toEpochSecond() * 1000
     }
 }
--- a/app/src/main/java/com/example/tasky/agenda_details/presentation/utils/DateTimeHelper.kt
+++ b/app/src/main/java/com/example/tasky/agenda_details/presentation/utils/DateTimeHelper.kt
@@ -1,5 +1,6 @@
 package com.example.tasky.agenda_details.presentation.utils
 
+import java.text.SimpleDateFormat
 import java.time.Duration
 import java.time.Instant
 import java.time.LocalDate
@@ -8,6 +9,8 @@ import java.time.ZoneId
 import java.time.ZoneOffset
 import java.time.ZonedDateTime
 import java.time.format.DateTimeFormatter
+import java.util.Date
+import java.util.Locale
 
 object DateTimeHelper {
     fun getLocalDateFromEpoch(
@@ -42,5 +45,11 @@ object DateTimeHelper {
         val localDate = LocalDate.parse(date, dateFormatter)
         val zdt = ZonedDateTime.of(localDate, localTime, ZoneOffset.UTC)
         return zdt.toEpochSecond() * 1000
+    }
+
+    fun formatEpochToDateString(epochMillis: Long): String {
+        val dateFormat = SimpleDateFormat("MMM d, HH:mm", Locale.getDefault())
+        val date = Date(epochMillis)
+        return dateFormat.format(date)
     }
 }
--- a/app/src/main/java/com/example/tasky/feature_agenda/domain/model/AgendaResponse.kt
+++ b/app/src/main/java/com/example/tasky/feature_agenda/domain/model/AgendaResponse.kt
@@ -1,0 +1,17 @@
+package com.example.tasky.feature_agenda.domain.model
+
+import com.example.tasky.agenda_details.domain.model.EventResponse
+import com.example.tasky.agenda_details.domain.model.ReminderResponse
+import com.example.tasky.agenda_details.domain.model.TaskResponse
+
+data class AgendaResponse(
+    val events: List<EventResponse> = emptyList(),
+    val tasks: List<TaskResponse> = emptyList(),
+    val reminders: List<ReminderResponse> = emptyList()
+)
+
+data class SyncAgendaResponse(
+    val deletedEventIds: List<String> = emptyList(),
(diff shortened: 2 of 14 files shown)


app/src/main/java/com/example/tasky/agenda_details/presentation/utils/DateTimeHelper.kt
```diff
@@ -36,11 +39,17 @@ object DateTimeHelper {
...
         val dateFormatter = DateTimeFormatter.ofPattern("MMM dd yyyy")
         val localTime = LocalTime.parse(time)
         val localDate = LocalDate.parse(date, dateFormatter)
         val zdt = ZonedDateTime.of(localDate, localTime, ZoneOffset.UTC)
-        return zdt.toEpochSecond()
+        return zdt.toEpochSecond() * 1000
+    }
+
+    fun formatEpochToDateString(epochMillis: Long): String {
+        val dateFormat = SimpleDateFormat("MMM d, HH:mm", Locale.getDefault())
```
philipplackner, line 51: The more modern way to format dates is by using the Java LocalDateTime API, since `SimpleDateFormat` is not thread-safe (not an issue here, but as soon as you'd share this `sdf` instance across threads you can run into issues). This API is also quite simpler than what you have here to get this working with time zones and all:
```
val dateTime = ZonedDateTime.ofInstant(
    Instant.ofEpochMilli(millis),
    ZoneId.of("UTC")
)
val formattedDateTime = DateTimeFormatter
    .ofPattern("dd MMM yyyy")
    .format(dateTime)
```

app/src/main/java/com/example/tasky/feature_agenda/domain/model/AgendaResponse.kt
```diff
@@ -0,0 +1,17 @@
...
+
+import com.example.tasky.agenda_details.domain.model.EventResponse
+import com.example.tasky.agenda_details.domain.model.ReminderResponse
+import com.example.tasky.agenda_details.domain.model.TaskResponse
+
+data class AgendaResponse(
+    val events: List<EventResponse> = emptyList(),
+    val tasks: List<TaskResponse> = emptyList(),
+    val reminders: List<ReminderResponse> = emptyList()
+)
```
philipplackner, line 11: Few things:
1. While you're not violating the architectural dependencies with putting this class in domain, you indirectly couple your domain layer to implementation details. The implementation detail in this case would be the names of the fields of the response (since they have to match the JSON field names from the API). The goal of isolating the domain layer like we do is to avoid having to change domain, the core of our app, when implementation details change. However, if the implementation detail (the JSON field name) would change here to something else, you would have to make changes in your domain layer. Having such responses in domain is kind of a special case, since you don't really see the implementation details (e.g. in form of an import from a 3rd party lib), but they're still there. If you really need to have the result of a response available across the app, you'll need to work with two models with this architecture, one response model in data which then gets mapped to an equivalent domain model in domain. But in this case, read (2) for what I recommend instead.
2. Right now, you'll be able to extract the agenda items out of this response object and use that in your app. But, you do already have a domain model that represents an agenda item (the `AgendaItem` class). So instead of making this response available to domain, you could map it to a `List<AgendaItem>` in your data layer.

app/src/main/java/com/example/tasky/feature_agenda/domain/model/AgendaResponse.kt
```diff
@@ -0,0 +1,17 @@
...
+
+data class SyncAgendaResponse(
+    val deletedEventIds: List<String> = emptyList(),
+    val deletedTasksIds: List<String> = emptyList(),
+    val deletedReminderIds: List<String> = emptyList()
+)
```
philipplackner, line 17: For this response (if I'm not wrong it should actually be a request, right?) it's similar, since in the end outside of the data layer, there's no use of knowing which items were deleted locally and have to be synced. If there's an agenda item that you deleted locally and still have to sync, you'll have to keep a persistent reference to it anyways in your DB. And if you have that reference in your DB, you can just get it in your data layer as well without having to pass this information through all of the layers. Here some pseudo code what I mean:
```
val deletedEventIds = deletedEventDao.getAllLocallyDeletedEventIds()
val deletedTaskIds = ...
val deletedReminderIds = ...

// This object is now created in data directly, so you don't need this information in domain anymore
val syncRequest = SyncAgendaRequest(
    deletedEventIds = deletedEventIds,
    // ...
)
```

app/src/main/java/com/example/tasky/feature_agenda/presentation/Agenda.kt
```diff
@@ -164,6 +174,34 @@ fun AgendaContent(
...
+                                    is AgendaItem.Task, is AgendaItem.Reminder ->
+                                        formatTimeBasedOnEvent(
+                                            agendaItem.date,
+                                            null
+                                        )
+                                },
+                                cardType = agendaItem.cardType,
+                                isChecked = false
+                            )
+                        }
```
philipplackner, line 202: A quick and easy way to add animations when items are inserted/changed/reordered is to add unique keys to each lazy column item and by adding a simple modifier:
```
items(
    items = agendaItems,
    key = { it.id } // use a property you know is unique per item
) {
    AgendaCard(
        ...,
        modifier = Modifier.animateItem()
    )
}
```

app/src/main/java/com/example/tasky/feature_agenda/presentation/AgendaViewState.kt
```diff
@@ -1,23 +1,55 @@
...
+    val agendaItems: List<AgendaItem>? = null
+)
+
+sealed class AgendaItem(
+    open val title: String,
+    open val details: String,
+    open val date: Long,
+    open val isChecked: Boolean,
+    open val cardType: AgendaItemType
+) {
```
philipplackner, line 33: You put this model here in presentation which is all fine if you just need it to display something. But agenda items and what data they consist of are such a core concept of this app, which is also needed in all layers. That's always a strong hint that something should be a domain model instead, as I also mentioned in an earlier comment. 

I know it's hard to wrap your head around these different layers and models. What helps me is to think of the app from a complete non-technical perspective. If there was no UI you had to worry about, no libraries, no APIs or databases, but just think of the problem the app solves (in other words, it's domain) - would there be a use for this model? 
This app solves the problem that you won't forget about your events and tasks anymore. That sentence already contains the information that it must be about some sort of event or task. So without including any technical details in the description of the app, we can't describe it without mentioning these core concepts of agenda items. Therefore, this must be a domain model.

Different example: If you had a social media app, like Instagram, what could be valid domain models? One could be a `Post` model for example, which includes information what a post consists of. A `Post` in the domain of Instagram is a completely different model than a `Post` would be in the context of Twitter, since those are two different domains. That's again a strong indication towards a domain model.

PR #38, opened 2024-08-02
Task and Reminder/api (create) 

sdelator:
### **Newly Added**

- agenda api calls (create task, create reminder)
- sort agenda items by from date

### **UI Changes**
image

### DIFF
--- a/app/src/main/java/com/example/tasky/agenda_details/domain/repository/AgendaDetailsRemoteRepository.kt
+++ b/app/src/main/java/com/example/tasky/agenda_details/domain/repository/AgendaDetailsRemoteRepository.kt
@@ -2,6 +2,7 @@ package com.example.tasky.agenda_details.domain.repository
 
 import com.example.tasky.agenda_details.domain.model.EventDetails
 import com.example.tasky.agenda_details.domain.model.EventResponse
+import com.example.tasky.agenda_details.domain.model.TaskResponse
 import com.example.tasky.common.domain.Result
 import com.example.tasky.common.domain.error.DataError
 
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/example/tasky/agenda_details/domain/repository/AgendaDetailsRemoteRepository.kt
```diff
@@ -16,4 +18,12 @@ interface AgendaDetailsRemoteRepository {
     suspend fun deleteEvent(): Result<Unit, DataError.Network>
 
     suspend fun updateEvent(): Result<EventResponse, DataError.Network>
+
+    suspend fun createTask(
+        taskDetails: TaskResponse
```
philipplackner, line 23: To extend my previous PR's feedback, the response here should also just be in data and you pass the domain model of `AgendaItem.Task` instead 👍 

If your model ends with Response, Request, DTO, Entity or similar, that's 100% a data model since it just represents how data is transferred/persistently saved

app/src/main/java/com/example/tasky/feature_agenda/presentation/AgendaViewModel.kt
```diff
@@ -70,20 +70,44 @@ class AgendaViewModel @Inject constructor(
...
+                                isChecked = false // todo updated by ui
+                            )
+                        })
+                    }
+                    val agendaSortedByTime = agendaItemList.sortedBy { agendaItem ->
+                        when (agendaItem) {
+                            is AgendaItem.Event -> agendaItem.fromDate
+                            is AgendaItem.Task -> agendaItem.date
+                            is AgendaItem.Reminder -> agendaItem.date
                         }
```
philipplackner, line 105: Given that all agenda item types have a `startDate`, could it make sense to shared this property to avoid these type checks?

philipplackner (approved):

PR #39, opened 2024-08-02
Data Class fix - Fixes for previous PR suggestions

sdelator:
-  map it to a List<AgendaItem> in your data layer

### DIFF
--- a/app/src/main/java/com/example/tasky/feature_agenda/data/repository/AgendaRemoteRepositoryImpl.kt
+++ b/app/src/main/java/com/example/tasky/feature_agenda/data/repository/AgendaRemoteRepositoryImpl.kt
@@ -6,7 +6,6 @@ import com.example.tasky.common.domain.Result
 import com.example.tasky.common.domain.error.DataError
 import com.example.tasky.common.domain.util.toNetworkErrorType
 import com.example.tasky.di.AuthenticatedApi
-import com.example.tasky.feature_agenda.domain.model.SyncAgendaResponse
 import com.example.tasky.feature_agenda.domain.repository.AgendaRemoteRepository
 import java.io.IOException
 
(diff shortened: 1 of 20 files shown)


app/src/main/java/com/example/tasky/feature_agenda/data/repository/AgendaRemoteRepositoryImpl.kt
```diff
@@ -1,30 +1,65 @@
...
                 val responseBody = response.body()
-                Result.Success(
-                    AgendaResponse(
-                        events = responseBody?.events ?: emptyList(),
-                        tasks = responseBody?.tasks ?: emptyList(),
-                        reminders = responseBody?.reminders ?: emptyList()
-                    )
-                )
+                if (responseBody != null) {
+                    val agendaItemList: List<AgendaItem> = mutableListOf<AgendaItem>().apply {
```
philipplackner, line 21: Not a big deal, but you can also quickly combine read-only lists without having to convert them to a mutable one:
```
val list1 = listOf(...)
val list2 = listOf(...)
val list3 = listOf(...)

return list1 + list2 + list3
```

app/src/main/java/com/example/tasky/feature_agenda/domain/model/AgendaResponse.kt
```diff
@@ -1,17 +1,11 @@
...
-    val reminders: List<ReminderResponse> = emptyList()
-)
-
-data class SyncAgendaResponse(
-    val deletedEventIds: List<String> = emptyList(),
-    val deletedTasksIds: List<String> = emptyList(),
-    val deletedReminderIds: List<String> = emptyList()
+    val events: List<Event> = emptyList(),
+    val tasks: List<Task> = emptyList(),
+    val reminders: List<Reminder> = emptyList()
```
philipplackner, line 10: This class is not needed in domain anymore. You just need it in data to receive and parse the response from the API, but then the rest of the app can just work with a `List<AgendaItem>`

app/src/main/java/com/example/tasky/feature_agenda/data/repository/AgendaRemoteRepositoryImpl.kt
sdelator: good idea! since they are read-only

PR #41, opened 2024-08-03
Task and Reminder/ create-api & Data class fix ! 

sdelator:
**### Newly Added**

- api call for create task and create reminder
- move agenda response to data layer 
- move events, reminder, and task data objects out of domain layer and into data layer
- create DTO mapper functions

### DIFF
--- a/app/src/main/java/com/example/tasky/agenda_details/presentation/AgendaDetailsViewModel.kt
+++ b/app/src/main/java/com/example/tasky/agenda_details/presentation/AgendaDetailsViewModel.kt
@@ -7,11 +7,14 @@ import com.example.tasky.agenda_details.domain.model.AgendaItem
 import com.example.tasky.agenda_details.domain.model.Attendee
 import com.example.tasky.agenda_details.domain.model.EventDetails
 import com.example.tasky.agenda_details.domain.model.Photo
+import com.example.tasky.agenda_details.domain.model.Reminder
+import com.example.tasky.agenda_details.domain.model.Task
 import com.example.tasky.agenda_details.domain.repository.AgendaDetailsRemoteRepository
 import com.example.tasky.agenda_details.presentation.utils.DateTimeHelper
 import com.example.tasky.common.domain.Result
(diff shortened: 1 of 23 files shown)


app/src/main/java/com/example/tasky/agenda_details/presentation/AgendaDetailsViewModel.kt
```diff
@@ -145,12 +146,101 @@ class AgendaDetailsViewModel @Inject constructor(
         //set viewstate to editing
     }
 
-    fun delete() {
+    fun delete(agendaItemType: AgendaItemType) {
         //todo delete functionality; navigate back to agenda
     }
 
-    fun save() {
+    fun save(agendaItemType: AgendaItemType) {
```
philipplackner, line 153: Isn't this an attribute you have available globally in this ViewModel? In that case, you wouldn't need to pass it as a parameter. Should just be a navigation argument, which you could retrieve using SavedStateHandle:
```
@HiltViewModel
class AgendaDetailsViewModel @Inject constructor(
    private val savedStateHandle: SavedStateHandle,
    ...
): ViewModel() {

    // Assuming agendaItemType is passed as a string navigation argument
    private val agendaItemType = savedStateHandle.get<String>("agendaItemType")?.let {
        AgendaItemType.valueOf(it)
    }
}
```

app/src/main/java/com/example/tasky/feature_agenda/data/repository/AgendaRemoteRepositoryImpl.kt
```diff
@@ -75,41 +76,41 @@ class AgendaRemoteRepositoryImpl(
...
+                        AgendaItem.Reminder(
+                            id = reminder.id,
+                            title = reminder.title,
+                            description = reminder.description,
+                            time = reminder.time,
+                            remindAt = reminder.remindAt
+                        )
                     }
+                    val agendaItemList = events + tasks + reminders
+                    //todo do we want to sort here by time ?
```
philipplackner, line 113: yep :D

PR #42, opened 2024-08-04
Various Small fixes 

sdelator:
- Fix Attendees
- remove SimpleDateFormat
- fix time conversions so that we store in UTC but display in user local time zone (PST)
- animate cards when loading


philipplackner (commented):
✅

PR #43, opened 2024-08-04
Event-Task-Reminder/api-delete

sdelator:
### **Newly Added**

- api call for delete event, task, and reminder

### DIFF
--- a/app/src/main/java/com/example/tasky/agenda_details/data/repository/AgendaDetailsRemoteRepositoryImpl.kt
+++ b/app/src/main/java/com/example/tasky/agenda_details/data/repository/AgendaDetailsRemoteRepositoryImpl.kt
@@ -113,12 +113,54 @@ class AgendaDetailsRemoteRepositoryImpl(
         }
     }
 
+    override suspend fun deleteTask(taskId: String): Result<Unit, DataError.Network> {
+        return try {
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/example/tasky/agenda_details/data/repository/AgendaDetailsRemoteRepositoryImpl.kt
```diff
@@ -113,6 +130,27 @@ class AgendaDetailsRemoteRepositoryImpl(
...
+            val response = api.deleteTask(taskId = taskId)
+
+            if (response.isSuccessful) {
+                val responseBody = response.body()
+                if (responseBody != null) {
+                    Result.Success(Unit)
+                } else {
+                    Log.e("Error", "API call failed with code ${response.code()}")
+                    Result.Error(DataError.Network.SERVER_ERROR)
+                }
```
philipplackner, line 144: Did you test this? I'm not sure this behaves as expected 🤔 
Just because an API doesn't respond with a body, doesn't mean the request failed. If the status code starts with a 2, the request was successful (which is already checked by `response.isSuccessful`)

sdelator: you are right ! will fix this

PR #45, opened 2024-08-05
Api/Get task, Get reminder, and Get event

sdelator:
**### Newly Added**

- add load reminder, load task api calls
- fix time conversions

### **UI Changes**

### DIFF
--- a/app/src/main/java/com/example/tasky/agenda_details/presentation/AgendaDetailsViewModel.kt
+++ b/app/src/main/java/com/example/tasky/agenda_details/presentation/AgendaDetailsViewModel.kt
@@ -11,6 +11,7 @@ import com.example.tasky.agenda_details.domain.repository.AgendaDetailsRemoteRep
 import com.example.tasky.agenda_details.presentation.utils.DateTimeHelper
 import com.example.tasky.common.domain.Result
 import com.example.tasky.common.domain.model.AgendaItemType
+import com.example.tasky.common.domain.util.convertMillisToMmmDdYyyy
 import com.example.tasky.common.presentation.CardAction
 import com.example.tasky.common.presentation.LineItemType
 import com.example.tasky.common.presentation.ReminderTime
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/example/tasky/agenda_details/presentation/AgendaDetailsViewModel.kt
```diff
@@ -91,15 +93,122 @@ class AgendaDetailsViewModel @Inject constructor(
     }
 
     private fun loadDataForEvent() {
-        // todo
+        if (agendaItemId == null) {
+            throw IllegalArgumentException("agendaItemId is null")
+        }
```
philipplackner, line 98: Couldn't this be null for a new agenda item or how do you check that?

app/src/main/java/com/example/tasky/feature_agenda/presentation/AgendaViewModel.kt
```diff
@@ -212,10 +212,10 @@ class AgendaViewModel @Inject constructor(
     fun formatTimeOnAgendaCard(fromDate: Long, toDate: Long? = null): String {
         println("fromDate $fromDate")
 
-        val startDate = DateTimeHelper.formatEpochToDateString(fromDate)
+        val startDate = fromDate.convertMillisToMMMdHHmm()
```
philipplackner, line 215: Instead of having a function that directly contains the Long timestamp to the formatted date/time, I'd recommend working with `ZonedDateTime` objects in your domain model instead (which you can then format in the UI). The advantage is that this makes it a lot easier to deal with times and dates, e.g. to add a few days, get time between two dates, etc. (whatever you may need, but it's just much easier to work with ZonedDateTime than with a Long timestamp)

app/src/main/java/com/example/tasky/agenda_details/presentation/AgendaDetailsViewModel.kt
sdelator: new agenda item does not call loadDataForEvent because its new :D

PR #46, opened 2024-08-05
Fix Agenda card dropdown - Fixes for previous PR suggestions

sdelator:
- anchoring dropdown to ellipsis image 

**### **UI Updates****
image
image

### DIFF
--- a/app/src/main/java/com/example/tasky/common/presentation/AgendaCard.kt
+++ b/app/src/main/java/com/example/tasky/common/presentation/AgendaCard.kt
@@ -37,7 +37,9 @@ fun AgendaCard(
     isChecked: Boolean,
     toggleAgendaCardDropdownVisibility: () -> Unit,
     showAgendaCardDropdown: Boolean,
-    onAgendaCardActionClick: (AgendaItem, CardAction) -> Unit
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/example/tasky/common/presentation/AgendaCard.kt
```diff
@@ -108,12 +120,31 @@ fun CustomCheckbox(isChecked: Boolean, color: Color, size: Dp) {
...
+    toggleAgendaCardDropdownVisibility: () -> Unit,
+    agendaItem: AgendaItem,
+    showAgendaCardDropdown: Boolean,
+    onAgendaCardActionClick: (AgendaItem, CardAction) -> Unit,
+    visibleAgendaCardDropdownId: String?,
+    setVisibleAgendaItemId: (String) -> Unit
+) {
+    Box {
+        IconButton(onClick = {
+            setVisibleAgendaItemId(agendaItem.id)
```
philipplackner, line 134: Is this extra function really needed? After all, you only care about when the drop down was toggled. You do have access to `agendaItem.id` in the parent composable of this as well, so you don't need to explicitly pass it.

PR #47, opened 2024-08-05
Api/update event, task, and reminder 

sdelator:
**### Newly Added**
- update event, task, reminder
- fix toolbar 

### **UI Change**

### DIFF
--- a/app/src/main/java/com/example/tasky/agenda_details/data/repository/AgendaDetailsRemoteRepositoryImpl.kt
+++ b/app/src/main/java/com/example/tasky/agenda_details/data/repository/AgendaDetailsRemoteRepositoryImpl.kt
@@ -101,24 +101,25 @@ class AgendaDetailsRemoteRepositoryImpl(
 
     override suspend fun updateEvent(
         eventDetails: EventDetailsUpdated,
-        photosByteArray: List<ByteArray>
+        photos: List<Photo.LocalPhoto>
     ): Result<AgendaItem.Event, DataError.Network> {
         return try {
             val gson = Gson()
             val json = gson.toJson(eventDetails)
             val requestBody = json.toRequestBody("application/json; charset=utf-8".toMediaType())
 
-            val photoParts = photosByteArray.mapIndexed { index, byteArray ->
-                val mimeType = when (byteArray.take(3)) {
-                    listOf(0xFF.toByte(), 0xD8.toByte(), 0xFF.toByte()) -> "image/jpeg"
-                    listOf(0x89.toByte(), 0x50.toByte(), 0x4e.toByte()) -> "image/png"
-                    else -> throw Exception("invalid Image format")
-                }
-                val requestFile =
-                    byteArray.toRequestBody(mimeType.toMediaTypeOrNull(), 0, byteArray.size)
+            val photoParts = photos.mapIndexed { index, localPhoto ->
+                val mimeTypeMap = MimeTypeMap.getSingleton()
+                val type = contentResolver.getType(Uri.parse(localPhoto.uri))
+                val extension = mimeTypeMap.getExtensionFromMimeType(type)
+                val requestFile = localPhoto.byteArray.toRequestBody(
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/example/tasky/agenda_details/data/repository/AgendaDetailsRemoteRepositoryImpl.kt
```diff
@@ -21,28 +26,30 @@ import okhttp3.RequestBody.Companion.toRequestBody
...
-                    listOf(0xFF.toByte(), 0xD8.toByte(), 0xFF.toByte()) -> "image/jpeg"
-                    listOf(0x89.toByte(), 0x50.toByte(), 0x4e.toByte()) -> "image/png"
-                    else -> throw Exception("invalid Image format")
-                }
-                val requestFile =
-                    byteArray.toRequestBody(mimeType.toMediaTypeOrNull(), 0, byteArray.size)
+            val photoParts = photos.mapIndexed { index, localPhoto ->
+                val mimeTypeMap = MimeTypeMap.getSingleton()
+                val type = contentResolver.getType(Uri.parse(localPhoto.uri))
+                val extension = mimeTypeMap.getExtensionFromMimeType(type)
```
philipplackner, line 44: Strictly speaking, you're violating the single responsibility principle here, since your function has multiple responsibilities (or in other words: multiple reasons to change):
1. The event update logic
2. How mime types are parsed

You don't have to be infinitely strict with that, it has to fit to the project's scale, however here I'd say putting this mime type conversion in a separate class would make sense to keep this reusable, since you probably also need this for creating events.

app/src/main/java/com/example/tasky/agenda_details/data/repository/AgendaDetailsRemoteRepositoryImpl.kt
```diff
@@ -92,8 +99,43 @@ class AgendaDetailsRemoteRepositoryImpl(
...
+    ): Result<AgendaItem.Event, DataError.Network> {
+        return try {
+            val gson = Gson()
+            val json = gson.toJson(eventDetails)
+            val requestBody = json.toRequestBody("application/json; charset=utf-8".toMediaType())
+
+            val photoParts = photos.mapIndexed { index, localPhoto ->
+                val mimeTypeMap = MimeTypeMap.getSingleton()
+                val type = contentResolver.getType(Uri.parse(localPhoto.uri))
+                val extension = mimeTypeMap.getExtensionFromMimeType(type)
```
philipplackner, line 114: Ah, I called it 😁

app/src/main/java/com/example/tasky/agenda_details/presentation/AgendaDetailsViewModel.kt
```diff
@@ -334,19 +339,165 @@ class AgendaDetailsViewModel @Inject constructor(
...
-    fun save() {
+    private fun updateEvent() {
+        val eventId = agendaItemId ?: throw IllegalArgumentException("agendaItemId is null")
+        val fromInEpochSeconds =
+            DateTimeHelper.getEpochMillisecondsFromDateAndTime(
+                date = _viewState.value.fromDate,
+                time = _viewState.value.fromTime
+            )
+        val toInEpochSeconds =
+            DateTimeHelper.getEpochMillisecondsFromDateAndTime(
```
philipplackner, line 350: What now, epoch seconds or milliseconds? The variable says something different than the function

sdelator: will fix ! needs to be ms

app/src/main/java/com/example/tasky/agenda_details/data/repository/AgendaDetailsRemoteRepositoryImpl.kt
sdelator: yes good idea! to clean up and make function readable

PR #48, opened 2024-08-06
Visitor Dialog- UI

sdelator:
### **### Newly Added**

- create UI for visitor Dialog
- check email if its valid

### **### UI Change**
image
image
image

### DIFF
--- a/app/src/main/java/com/example/tasky/agenda_details/presentation/AgendaDetailsRoot.kt
+++ b/app/src/main/java/com/example/tasky/agenda_details/presentation/AgendaDetailsRoot.kt
@@ -328,6 +328,7 @@ fun AgendaDetailsContent(
                 GrayDivider()
                 if (agendaItemType == AgendaItemType.Event) {
                     AttendeeSection(
+                        isEditMode = isEditMode,
                         attendeeFilter = attendeeFilter,
                         onAttendeeFilterClick = onAttendeeFilterClick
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/example/tasky/agenda_details/presentation/AgendaDetailsRoot.kt
```diff
@@ -75,6 +75,8 @@ fun AgendaDetailsRoot(
     agendaDetailsViewModel: AgendaDetailsViewModel = hiltViewModel()
 ) {
     val viewState by agendaDetailsViewModel.viewState.collectAsStateWithLifecycle()
+    val email by agendaDetailsViewModel.email.collectAsStateWithLifecycle()
+    val isEmailValid by agendaDetailsViewModel.isEmailValid.collectAsStateWithLifecycle()
```
philipplackner, line 79: Similar question as I had earlier already - why aren't these part of your viewState? The email and whether it's valid definitely has an impact on the UI, right?

philipplackner (approved):

app/src/main/java/com/example/tasky/agenda_details/presentation/AgendaDetailsRoot.kt
sdelator: fixing this !!

PR #49, opened 2024-08-06
Get Attendee and Delete attendee/api 

sdelator:
### **Newly Added** 

- attendee api call and delete function
- if user does not exist, show user is not valid underneath the text field
- set user as default Going in list

### **UI Change**
image
image
image


philipplackner (commented):
✅

PR #50, opened 2024-08-07
 Fixes for previous PR suggestions

sdelator:

- add all variables to viewstate 
- practice single responsibility principle for function with Mime
- Fix epoch to millisecond conversions
- Use ZonedDateTime in domain

### DIFF
--- a/app/src/main/java/com/example/tasky/agenda_details/data/MimeUtils.kt
+++ b/app/src/main/java/com/example/tasky/agenda_details/data/MimeUtils.kt
@@ -1,0 +1,23 @@
+package com.example.tasky.agenda_details.data
+
+import android.content.ContentResolver
+import android.net.Uri
+import android.webkit.MimeTypeMap
+import okhttp3.MediaType.Companion.toMediaTypeOrNull
+import okhttp3.RequestBody
+import okhttp3.RequestBody.Companion.toRequestBody
+
+object MimeUtils {
+    fun getMimeType(contentResolver: ContentResolver, uri: String): String? {
+        val type = contentResolver.getType(Uri.parse(uri))
+        return type
+    }
+
+    fun getExtensionFromMimeType(mimeType: String?): String? {
+        return MimeTypeMap.getSingleton().getExtensionFromMimeType(mimeType)
+    }
+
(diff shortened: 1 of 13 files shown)


app/src/main/java/com/example/tasky/agenda_details/presentation/AgendaDetailsRoot.kt
```diff
@@ -378,7 +376,7 @@ fun AgendaDetailsContent(
                         )
                     }
                 }
-                item { Spacer(modifier = Modifier.weight(1f)) }
+                item { Spacer(modifier = Modifier.padding(100.dp)) }
```
sdelator, line 379: weight does not work inside of lazyColumn :(

app/src/main/java/com/example/tasky/agenda_details/data/MimeUtils.kt
```diff
@@ -0,0 +1,23 @@
...
+        return type
+    }
+
+    fun getExtensionFromMimeType(mimeType: String?): String? {
+        return MimeTypeMap.getSingleton().getExtensionFromMimeType(mimeType)
+    }
+
+    fun toRequestBody(byteArray: ByteArray, mimeType: String?): RequestBody {
+        return byteArray.toRequestBody(mimeType?.toMediaTypeOrNull())
+    }
```
philipplackner, line 22: This function is not really the job of a MimeUtils class. MimeTypes themselves have nothing to do with request bodies, that's just how you use the mime type in this specific case. Instead, I'd recommend to create the request body in your repository while using the functions from this MimeUtils class there.

app/src/main/java/com/example/tasky/agenda_details/presentation/AgendaDetailsRoot.kt
philipplackner: Yes, since a lazy column doesn't know all of its items (some of these might not be loaded due to the lazy behavior), it also can't calculate the dimensions of them all. Simple padding will be fine here. If you'd really like to have the weight behavior, you could use the lazy list state to check if it's scrollable and if not, calculate the remaining height of the screen after subtracting all visible elements to assign that for the spacer.

philipplackner (approved):

PR #52, opened 2024-08-08
Notifications 

sdelator:
**### Newly Added**
- create notification service for all agenda items
- update notifications when items get updated
- delete notification and alarm when items get deleted
- create Worker to create and display notifications
- user alarm manager to schedule

**### UI Change**
image


app/src/main/AndroidManifest.xml
```diff
@@ -24,6 +28,7 @@
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
         </activity>
+        <receiver android:name=".common.data.notification_service.NotificationBroadcastReceiver" />
```
philipplackner, line 31: It's also important to listen to when the Android device boots up, since all AlarmManager reminders will be cleared when the device is turned off, so you'll have to reschedule

app/src/main/java/com/example/tasky/common/data/notification_service/NotificationHandlerImpl.kt
```diff
@@ -0,0 +1,109 @@
...
+        notificationId: String,
+        title: String,
+        description: String,
+        remindAt: Long
+    ) {
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.S) {
+            val alarmManager = context.getSystemService(Context.ALARM_SERVICE) as AlarmManager
+            if (!alarmManager.canScheduleExactAlarms()) {
+                // Request permission to schedule exact alarms
+                val intent = Intent(Settings.ACTION_REQUEST_SCHEDULE_EXACT_ALARM)
```
philipplackner, line 54: For an app like Tasky, I'd argue you could also use the `USE_EXACT_ALARM` permission which doesn't require the users approval. Google is stricter with that one when publishing an app, but if scheduling exact alarms makes up a core feature of your app (which it definitely is for a calendar app), they'll grant it.

app/src/main/java/com/example/tasky/common/data/notification_service/NotificationBroadcastReceiver.kt
```diff
@@ -0,0 +1,42 @@
...
+
+        val data = Data.Builder()
+            .putInt("notificationId", notificationId)
+            .putString("title", title)
+            .putString("description", description)
+            .build()
+
+        val workRequest = OneTimeWorkRequest.Builder(ReminderNotificationWorker::class.java)
+            .setInputData(data)
+            .build()
```
philipplackner, line 38: I wonder why you use WorkManager to show the notification? That's not a long running task, so you could just use the NotificationManager here and call `notify(...)` with the data you got

PR #53, opened 2024-08-08
Notification Navigation

sdelator:
**### Newly Added**
- navigate to corresponding detail screen after clicking notification

### DIFF
--- a/app/src/main/java/com/example/tasky/common/data/notification_service/ReminderNotificationWorker.kt
+++ b/app/src/main/java/com/example/tasky/common/data/notification_service/ReminderNotificationWorker.kt
@@ -6,11 +6,12 @@ import android.app.NotificationManager
 import android.app.PendingIntent
 import android.content.Context
 import android.content.Intent
+import android.net.Uri
 import androidx.core.app.NotificationCompat
 import androidx.work.Worker
 import androidx.work.WorkerParameters
-import com.example.tasky.MainActivity
 import com.example.tasky.R
+import com.example.tasky.common.presentation.CardAction
 
 /**
  * Handles the actual work of creating and displaying the notification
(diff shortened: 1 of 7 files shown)


app/src/main/java/com/example/tasky/common/data/notification_service/ReminderNotificationWorker.kt
```diff
@@ -51,13 +70,18 @@ class ReminderNotificationWorker(
...
         manager?.createNotificationChannel(channel)
         //todo add deepLinkNavigationData
-        val activityIntent =
-            Intent(context, MainActivity::class.java) //todo navigate to specific screen
+        val cardAction = CardAction.Open.name
+        val activityIntent = Intent(Intent.ACTION_VIEW).apply {
+            data =
+                Uri.parse("taskyapp://details/$date?agendaItemType=$agendaItemType&agendaItemId=$agendaItemId&cardAction=$cardAction")
+            addFlags(Intent.FLAG_ACTIVITY_NEW_TASK or Intent.FLAG_ACTIVITY_CLEAR_TASK)
+        }
```
philipplackner, line 78: By doing it like this, it could be that you mess up the back stack. I'm not 100% sure at this point, but have you tried what happens when you visit the detail screen from this deep link and then press back? It might close the app or lead to a similar unexpected result, even though users would expect to get to the agenda.

To construct a deep link Intent, this is the typical way to do that while preserving the back stack:
```
val activityIntent = Intent(app, MainActivity::class.java).apply {
    // Construct intent without the flags
}

val pendingIntent = TaskStackBuilder.create(app).run {
    addNextIntentWithParentStack(activityIntent) // <- Important
    getPendingIntent(
        item.hashCode(),
        PendingIntent.FLAG_UPDATE_CURRENT or PendingIntent.FLAT_IMMUTABLE
    )
}
```

PR #54, opened 2024-08-08
Local Database - add, update, delete

sdelator:
**### Newly Added**
- (Add/Update/Delete) event, reminder, task to local database

### **UI Changes**
image
image

### DIFF
--- a/app/src/main/java/com/example/tasky/agenda_details/presentation/AgendaDetailsViewModel.kt
+++ b/app/src/main/java/com/example/tasky/agenda_details/presentation/AgendaDetailsViewModel.kt
@@ -328,6 +328,11 @@ class AgendaDetailsViewModel @Inject constructor(
                     }
                 }
             }
+
+            withContext(Dispatchers.IO) {
+                // delete in local db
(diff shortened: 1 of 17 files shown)


app/src/main/java/com/example/tasky/agenda_details/presentation/AgendaDetailsViewModel.kt
```diff
@@ -317,6 +326,11 @@ class AgendaDetailsViewModel @Inject constructor(
                     }
                 }
             }
+
+            withContext(Dispatchers.IO) {
+                // delete in local db
+                eventLocalRepository.deleteEvent(eventId = eventId)
+            }
```
philipplackner, line 333: The switch here is not necessary. The responsibility to switch to the right dispatcher lies at the function where the blocking call gets converted to the suspending function. In this case, somewhere deep down in the Room library there will be a blocking (and not suspending) call which wraps the blocking call with `withContext(Dispatchers.IO)`.

Everything else would be super error-prone since not remembering to switch to the right dispatcher would always result in worse performance or bugs.

PR #55, opened 2024-08-09
Needle

sdelator:
**### Newly Added**
- add agenda needle for current time
- Fix strikethrough/checkmark when Task is checked 

**### UI Changes**
image

### DIFF
--- a/app/src/main/java/com/example/tasky/common/presentation/AgendaCard.kt
+++ b/app/src/main/java/com/example/tasky/common/presentation/AgendaCard.kt
@@ -20,7 +20,9 @@ import androidx.compose.ui.res.colorResource
 import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.text.TextStyle
+import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.text.style.TextDecoration
 import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.dp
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/example/tasky/common/presentation/AgendaCard.kt
```diff
@@ -103,10 +118,25 @@ fun AgendaCard(
 }
 
 @Composable
-fun CustomCheckbox(isChecked: Boolean, color: Color, size: Dp) {
-    IconButton(onClick = { /*TODO*/ }) {
+fun CustomCheckbox(
```
philipplackner, line 121: I'd never name Composables starting with `Custom`, since that doesn't tell the caller anything about its purpose and in how far it's custom. In the end, every Composable is custom, isn't it? 😄  
If you're creating a Composable that represents a default styled component used across the project like here, I like to communicate that with the name instead. For example, `TaskyTextField`. That at least communicates that it's a component with the default style from the mockups which should be used for all components of this type across the project.

app/src/main/java/com/example/tasky/common/presentation/AgendaCard.kt
```diff
@@ -103,10 +118,25 @@ fun AgendaCard(
...
+    taskCompleteForAgendaItemId: String = ""
+) {
+    IconButton(onClick = {
+        if (agendaItem != null) {
+            if (agendaItem.cardType == AgendaItemType.Task) {
+                setTaskCompleteForAgendaItemId(agendaItem.id)
+                toggleTaskComplete()
+            }
+        }
+    }) {
```
philipplackner, line 137: Instead of checking for the type in onClick, I'd probably rather set this button disabled for events and reminders, so you don't get the ripple there. Otherwise it might raise the impression that you could click on it, but it doesn't work

app/src/main/java/com/example/tasky/feature_agenda/presentation/Agenda.kt
```diff
@@ -260,10 +270,15 @@ fun AgendaContent(
...
-                        items(
-                            items = agendaItems,
-                            key = { it.id }
-                        ) { agendaItem ->
+                        itemsIndexed(
+                            items = agendaItems
+                        ) { index, agendaItem ->
+                            if (index > 0 && agendaItem.startTime.toLocalDate() == needlePosition.toLocalDate() && agendaItem.startTime.isAfter(
+                                    needlePosition
+                                ) && agendaItems[index - 1].startTime.isBefore(needlePosition)
```
philipplackner, line 278: What if all agenda items are later than now? Then the needle would have to show up as the first element, but the first `index > 0` check already prevents this or am I wrong?

app/src/main/java/com/example/tasky/feature_agenda/presentation/Agenda.kt
```diff
@@ -260,10 +270,15 @@ fun AgendaContent(
...
+                            ) {
+                                Needle()
```
philipplackner, line 280: I personally created a separate class that treats the Needle as its own element in the list:
```
sealed interface AgendaItemUi {
    data class Item(val item: AgendaItem) : AgendaItemUi
    data object Needle : AgendaItemUi
}
```
Then the UI will get a List<AgendaItemUi> and the ViewModel can handle the logic to decide where the needle is. This has the advantage that you can easily animate item changes, e.g. when time passes and the needle moves past the next item.

